PR #145, opened 2026-04-02
run restore on all stf periodically

sapinb:
## Description
Run stf::restore() periodically on all state machines based on configured time interval.

This is a catch-all to resolve any lost actions that prevent the stf from continuing execution. 
Equivalent to periodically restarting the mosaic node.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency update
- [ ] Security fix

## Notes to Reviewers
Job scheduler does not currently dedupe actions. With periodic restores, this can potentially cause duplicate job runs, although it will not impact the correctness of the process, just some inefficiency.

## Checklist

### DIFF
--- a/bin/mosaic/config/config.a.toml
+++ b/bin/mosaic/config/config.a.toml
@@ -62,6 +62,7 @@ chunk_timeout_secs = 30
 
 [sm_executor]
 command_queue_size = 256
+restore_interval_secs = 60
 
 [rpc]
 bind_addr = "127.0.0.1:8000"
--- a/bin/mosaic/config/config.b.toml
+++ b/bin/mosaic/config/config.b.toml
@@ -62,6 +62,7 @@ chunk_timeout_secs = 30
 
 [sm_executor]
 command_queue_size = 256
+restore_interval_secs = 60
 
 [rpc]
 bind_addr = "127.0.0.1:8001"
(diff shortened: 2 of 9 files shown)


AaronFeickert (approved):

Zk2u (changes requested):
If we are going to do this, we should clear the job queue then restore into it. This needs to ensure that the clearing operation is done properly, doesn't have race conditions or violate any of the other guarantees the job scheduler makes to the other components

Zk2u:
Blocking issue in `crates/sm-executor/src/lib.rs` inside `run_inner`.

This PR still rebuilds `self.job_handle.recv()`, `self.net_client.recv()`, and `self.command_rx.recv()` directly inside `monoio::select!` on every loop iteration. That is the cancel-unsafe pattern we just fixed in `#167`: if another branch wins after a direct handoff has matched a waiting `kanal` receiver, the dropped receive future can drop the delivered message.

So as written, this PR reintroduces the same completion-loss class we traced for the `GeneratingPolynomialCommitments` wedge. Before this can merge, it needs to be rebased onto current `main` and the periodic restore tick integrated with the persistent pinned receive futures pattern from `#167`, rather than the old per-iteration `recv()` futures.

This is on the assumption this is actually a good idea, which i'm not sure it is and we should discuss on slack first

AaronFeickert:
Moving to draft until a decision is made on the exact approach to take.

sapinb:
Retry related issues have been fixed at the job and state machine executor levels in #163 and #167 which cover all currently known issues that require retries. Closing this PR for now.

PR #163, opened 2026-04-08
sm-executor: stop dropping completed job results

### LINKED ISSUE #153: sm-executor: stop dropping completed job results on transient apply/commit failures
## Summary

A job can complete successfully and still never update the state machine.

Today, if `sm-executor` receives a completion and then hits a non-fatal error while applying it, it logs and drops the completion permanently. The same general problem exists if a successful completion cannot be delivered back through the completion channel. This means meaningful completed work can be lost after execution, which violates the core durability/correctness invariant of the job system.

This is also the most likely explanation for the original `GeneratingPolynomialCommitments` wedge where a small number of wire results were missing until restart/restore re-emitted the work.

This issue is about preserving the invariant that once meaningful work has completed, its result is not silently lost during delivery or application.

## Correct model

There are three distinct phases:

1. job execution
2. completion delivery
3. completion application to SM state

If phase 1 succeeded, the system must not lose the result during phases 2 or 3 just because storage, commit, or application hit a transient failure.

The correct behavior is:

1. **Transient execution failure before a completion exists**
- retry the job

2. **Transient failure while delivering or applying an already-produced completion**
- internally retry/requeue application of that completion inside `sm-executor`
- do not drop the completion

3. **Completed domain/protocol failure verdict**
- surface it through the normal completion/result model
- do not silently convert it into a dropped completion or infinite retry

The key point is that “work completed” and “state durably updated” are separate concerns, and the system currently treats some failures in the second category as if the result can be discarded.

## Desired solution

1. Audit all completion handling paths in `sm-executor` and classify which apply-time errors are transient and must be retried rather than dropped.
2. Implement internal retry/requeue for already-produced completions inside `sm-executor`.
3. Ensure storage/commit failures during completion application do not permanently wedge the protocol instance.
4. Ensure successful completions are not silently lost if a completion channel closes after the work has already finished.
5. Add explicit logging/metrics around:
- completion produced
- completion delivery failed
- completion application failed transiently
- completion requeued
- completion durably applied

The intended fix here is internal retry/requeue in `sm-executor`, not a separate external durability mechanism. Once a completion exists, `sm-executor` should keep trying to apply it until it either succeeds or hits a genuinely terminal condition.

## Scope of this issue

1. Audit completion handling in `sm-executor`.
2. Audit completion send/drop behavior in the regular worker pool and garbling coordinator.
3. Reclassify transient apply-time failures so they trigger internal requeue behavior rather than “warn and drop”.
4. Add tests covering completion delivery/application failure after successful job execution.

## Acceptance criteria

- A transient storage or commit failure during completion application cannot permanently wedge setup.
- Successful polynomial-commitment jobs are not lost due to apply-time failures.
- Restart/restore is no longer required to recover from transient completion-application failures.
- Successful job completions are not silently discarded if delivery/application temporarily fails.
- `sm-executor` internally requeues/retries failed completion application until success or terminal classification.
- Tests cover transient failure after successful job execution and prove eventual durable application.

Zk2u:
Closes #153.

## Summary

This change closes the completion-loss gaps behind #153 while keeping the scheduler/SM boundary aligned with the current design assumptions.

First, `sm-executor` no longer drops completed results when applying them fails transiently. Completed jobs are now queued locally and retried with backoff for `Storage` and `Commit` failures instead of being logged and discarded.

Second, completion-side STF rejections are no longer treated as infrastructure failures. If a worker completion is delivered successfully but the state machine rejects it (for example `UnexpectedInput` from a stale completion), `sm-executor` now logs and drops that completion instead of shutting down the executor.

Third, the scheduler no longer silently loses a completed result if the completion channel is closed. At that point the underlying job has already run and may already have caused protocol-visible side effects, so replaying it is unsafe. Instead, pool workers and garbling workers now signal an internal fatal scheduler fault, and the scheduler shuts down fail-closed.

## Why

The old behavior had two bad outcomes:

- `sm-executor` could lose a completion after the work had already finished, purely because applying or committing state failed transiently.
- the scheduler could lose a completion if delivery back to the SM path failed, and it would do so silently.

Those need different fixes:

- apply-time failure inside `sm-executor` is safe to retry because we are retrying completion application, not re-running the job
- completion-side STF rejection is an SM/protocol semantic problem, not a scheduler correctness failure, so it should be logged and dropped rather than crashing the executor
- delivery failure after the job already executed is not safe to replay blindly, because many jobs can already have produced network/protocol side effects

## What changed

- `sm-executor` now keeps a pending-completion queue with exponential backoff and retries `Storage` / `Commit` failures
- completion-side `SmExecutorError::Stf` now logs and drops the offending completion instead of stopping the executor
- dropped/requeued completion logs now include the action id for debugging
- `JobCompletion` and `ActionCompletion` are now `Clone` so completions can be retained across retries
- the job scheduler now has an internal `SchedulerFault` path
- pool workers and garbling workers report fatal completion-delivery failures instead of silently discarding completions
- the scheduler shuts down fail-closed when that happens
- formatted the repo TOML files that were blocking `taplo format --check`
- added regression tests for:
  - retry classification and requeue behavior in `sm-executor`
  - full executor-loop retry after a transient completion-application failure
  - drop behavior for completion-side STF errors
  - pool-worker completion-channel closure
  - garbling-worker completion-channel closure

## Validation

Passed:

- `just -f .justfile ci`
- `PATH="/usr/local/libexec:$PATH" ./run_tests.sh -t tests/fn_mosaic_setup.py`
- `cargo test -p mosaic-sm-executor`
- `cargo clippy -p mosaic-sm-executor --tests -- -D warnings`

The focused functional test still exposes the separate protocol bug tracked in #165, but it no longer fails by shutting down `sm-executor` on a stale completion.

## Reviewer notes

This PR intentionally does not try to recover from scheduler-side completion-channel closure without shutdown. Once the job has already completed, replaying it can duplicate side effects. In the current architecture, fail loud / fail closed is the safe behavior there.

The completion policy is now intentionally split three ways:

- retry transient apply failures (`Storage`, `Commit`)
- log/drop SM semantic rejections (`Stf`)
- fail closed on infrastructure/runtime breakage (`SourceClosed`, `NetRecv`, `StfPanic`, scheduler delivery failure, etc.)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -64,6 +64,7 @@ target/
 # Tests artifacts
 mutants.out*
 lcov.info
+functional-tests/_dd_parallel_shard*/
 
 # Unit test / coverage reports
 htmlcov/
--- a/crates/job/api/src/submission.rs
+++ b/crates/job/api/src/submission.rs
@@ -102,7 +102,7 @@ impl JobActions {
 ///
 /// The peer ID on [`JobCompletion`] plus the variant here (garbler vs
 /// evaluator) identifies which SM to deliver to.
-#[derive(Debug)]
+#[derive(Debug, Clone)]
 pub enum ActionCompletion {
     /// Garbler tracked action completed.
     Garbler {
@@ -180,7 +180,7 @@ impl ActionCompletion {
 ///
 /// Jobs always retry internally until they succeed — every submitted action
 /// eventually produces exactly one completion. There is no failure variant.
-#[derive(Debug)]
+#[derive(Debug, Clone)]
 pub struct JobCompletion {
     /// The peer whose SM this result should be routed to.
     pub peer_id: PeerId,
--- a/crates/job/scheduler/src/garbling/mod.rs
+++ b/crates/job/scheduler/src/garbling/mod.rs
@@ -58,6 +58,8 @@ use mosaic_job_api::{
 use mosaic_net_svc_api::PeerId;
 use tracing::Instrument;
 
+use crate::SchedulerFault;
+
 /// Size of each v5c gate record in bytes (3 × u32).
 const GATE_SIZE: usize = 12;
 
@@ -259,10 +261,11 @@ impl GarblingCoordinator {
     /// Jobs submitted via [`submit`](Self::submit) are collected into batches,
     /// sessions are created via the `factory`, and workers process them
     /// concurrently.
-    pub fn new(
+    pub(crate) fn new(
         config: GarblingConfig,
         factory: Arc<dyn SessionFactory>,
         completion_tx: kanal::AsyncSender<JobCompletion>,
+        fault_tx: kanal::AsyncSender<SchedulerFault>,
     ) -> Self {
         let (submit_tx, submit_rx) = kanal::bounded_async(config.max_concurrent * 2);
 
@@ -273,7 +276,13 @@ impl GarblingCoordinator {
                     .enable_timer()
                     .build()
                     .expect("failed to build monoio runtime for garbling coordinator")
-                    .block_on(coordinator_loop(config, factory, submit_rx, completion_tx));
+                    .block_on(coordinator_loop(
+                        config,
+                        factory,
+                        submit_rx,
+                        completion_tx,
+                        fault_tx,
+                    ));
             })
             .expect("failed to spawn garbling coordinator thread");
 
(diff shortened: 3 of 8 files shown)


sapinb (approved):
ACK

sapinb:
@Zk2u Please rebase over latest main and resolve conflicts

PR #167, opened 2026-04-08
Recover garbling jobs and make sm-executor receives cancel-safe

### LINKED ISSUE #156: job-scheduler: recover garbling sessions on worker/report failure instead of losing jobs
## Summary

The garbling coordinator currently has explicit failure paths where active sessions are lost if a worker dies, fails to report, or exits mid-pass. In those cases, the original `PendingCircuitJob`s are not reconstructed and requeued, so in-flight work can be lost permanently.

This issue is about making the coordinator the durable owner of in-flight circuit jobs so worker loss becomes recoverable rather than terminal.

## Correct model

A worker is an execution vehicle, not the owner of job durability.

If a worker disappears, times out, or stops reporting, the coordinator must still know which jobs were assigned to that worker and must be able to requeue them.

The correct behavior is:

1. **Worker/report failure**
- recover and requeue all affected in-flight jobs

2. **Transient session execution failure**
- return the job to pending retry

3. **True permanent setup/programming failure**
- classify explicitly and drop only if the error is genuinely permanent

The key point is that the coordinator must never move a job into a worker and then lose the ability to reconstruct it.

## Desired solution

1. Keep authoritative ownership of in-flight `PendingCircuitJob`s at the coordinator level even after assignment to workers.
2. Track which jobs are assigned to each worker for the current pass.
3. On assignment failure, missing chunk report, missing finish report, or worker exit:
- recover all sessions known to be on that worker
- requeue them into `pending_retry`
4. Audit and narrow `SetupFailed` classification so only truly permanent/setup-programming failures are dropped.
5. Add tests that simulate worker loss or report timeout mid-pass and verify all affected jobs are retried.

The implementation can choose the exact tracking structure, but it must no longer be possible for worker disappearance to imply unrecoverable job loss.

## Scope of this issue

1. Audit worker assignment and pass bookkeeping in the garbling coordinator.
2. Introduce recoverable tracking for assigned/in-flight circuit jobs.
3. Rework worker failure/report-timeout handling to requeue affected jobs.
4. Audit permanent-error classification for circuit session creation and finish paths.
5. Add failure-injection tests for assignment failure, missing chunk report, missing finish report, and worker exit with active sessions.

## Acceptance criteria

- Worker death or report timeout does not cause unrecoverable garbling job loss.
- All in-flight circuit jobs remain reconstructible by the coordinator.
- Existing “sessions are lost” paths are removed or replaced with retry/recovery behavior.
- Only genuinely permanent setup/programming failures are dropped.
- Tests cover the major worker/report failure modes and prove retry/recovery.

Zk2u:
## Summary

This PR addresses two independent reliability bugs in the scheduler/executor layer:

- fixes #156 by making the garbling coordinator retain authoritative ownership of in-flight `PendingCircuitJob`s and requeue them on assignment/report/worker failure
- fixes #166 by making `sm-executor` main-loop receives cancel-safe so completed jobs cannot be lost when another `select!` branch wins

### Garbling coordinator recovery

- the coordinator now tracks assigned jobs per worker instead of letting them become unrecoverable once handed off
- assignment failure, missing chunk report, missing finish report, and unexpected worker exit now requeue the affected jobs
- finish completions are forwarded back through the coordinator while preserving retry ownership
- `PendingCircuitJob` is clonable so coordinator-owned recovery bookkeeping stays explicit

### sm-executor cancel safety

- `sm-executor` no longer recreates `recv()` futures inside `monoio::select!` every loop iteration
- it now keeps persistent pinned receive futures for:
  - job completions
  - inbound network requests
  - executor commands
  - shutdown
- this avoids the proven `kanal` direct-handoff cancellation hole where a completion can be dropped if a waiting `recv()` future is canceled before being polled to completion

## Tests

Added/updated tests for:

- garbling coordinator recovery on assignment failure
- garbling coordinator recovery on missing chunk report
- garbling coordinator recovery on missing finish report
- finish report forwarding plus retry preservation
- `kanal` waiting-receive drop loses direct-handoff message
- persistent receive future preserves direct-handoff message

## Validation

Passed:

- `cargo test -p mosaic-sm-executor`
- `cargo clippy -p mosaic-sm-executor --tests -- -D warnings`
- `cargo test -p mosaic-job-scheduler`
- `cargo clippy -p mosaic-job-scheduler --tests -- -D warnings`
- `cargo test -p mosaic-job-api`
- `cargo fmt --check --all`

`just -f .justfile ci` got through formatting, clippy, docs, and the scheduler/executor portions cleanly, but timed out in unrelated long-running `mosaic-storage-fdb` tests in this environment.

## Notes

I also updated issue #156 to clarify that the rare `GeneratingPolynomialCommitments` wedge is tracked separately in #166 and is not part of the coordinator worker-loss bug.


sapinb (approved):
LGTM

PR #169, opened 2026-04-13
add S3 IRSA table store backend

sapinb:
## Description
Pass aws credentials using IAM roles for service accounts (IRSA)

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency update
- [ ] Security fix

## Checklist

- [ ] I have commented my code where necessary.

### DIFF
--- a/bin/mosaic/src/config.rs
+++ b/bin/mosaic/src/config.rs
@@ -306,12 +306,9 @@ pub(crate) enum TableStoreBackend {
         bucket: String,
         region: String,
         prefix: String,
-<<<<<<< HEAD
-=======
         /// When both `access_key_id` and `secret_access_key` are omitted, the
         /// AWS credential chain is used: IRSA web-identity token → ECS task
         /// creds → EC2 instance profile.
->>>>>>> 3e11546 (remove IRSA backend, allow optional access_key, secret in S3Compatible backend for IRSA)
         access_key_id: Option<String>,
         secret_access_key: Option<String>,
         endpoint: Option<String>,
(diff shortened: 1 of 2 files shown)


Zk2u (changes requested):
Requesting changes.

I think this should be modeled as part of the existing S3 backend rather than as a separate `S3Irsa` backend.

The implementation here does not introduce a different table store behavior. Both variants build the same `AmazonS3Builder`, the same `S3TableStore`, and run through the same startup path. The only real difference is credential sourcing: static credentials vs the default AWS credential chain.

That makes this look like a configuration mode of the same backend, not a distinct backend type.

Why I’m asking for a change:
- it duplicates validation and startup branching for what is still one S3/object-store implementation
- it creates avoidable drift if we add more shared S3 options later
- `S3Irsa` is also narrower than the actual behavior here, since `AmazonS3Builder::new()...build()` uses the broader default AWS credential chain, not just IRSA

I’d prefer one S3 backend with optional static credentials:
- if `access_key_id` / `secret_access_key` are provided, use them
- otherwise fall back to the default AWS credential chain

That keeps AWS S3, IRSA/web-identity, instance-profile, and static-credential usage under one backend shape, while still supporting endpoint/path-style options for compatible stores.

If you want a naming split, I think it would need to reflect a real behavioral/config boundary, not just credential source. As written, I don’t think `S3Irsa` earns a separate enum variant.

sapinb:
@Zk2u Removed the separate backend variant and just made the fields optional.

Zk2u (approved):

PR #170, opened 2026-04-13
build docker image with smaller cac parameters for testing

sapinb:
## Description
To reduce setup time during testing deployments, build circuit using smaller cac params (5/3 instead of 181/174)

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency update
- [ ] Security fix

## Checklist

- [ ] I have commented my code where necessary.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -8,9 +8,9 @@ FROM ubuntu:24.04 AS builder
 # Install build dependencies and Rust nightly toolchain.
 RUN apt-get update \
     && apt-get install -y --no-install-recommends \
-         adduser build-essential ca-certificates clang curl libclang-dev pkg-config \
+    adduser build-essential ca-certificates clang curl libclang-dev pkg-config \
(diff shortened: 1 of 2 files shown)


Zk2u:
Updated this so the Docker image can switch into reduced-circuits mode via env var instead of baking that choice in permanently.

What changed:
- the image now builds both binaries:
  - default `mosaic`
  - reduced `mosaic-reduced`
- the entrypoint selects the reduced binary when `MOSAIC_REDUCED_CIRCUITS` is set to a truthy value (`1`, `true`, `yes`, `on`)
- if the env var is unset, it runs the normal binary

That keeps one image usable for both normal and reduced-circuit testing without needing separate image variants.

Zk2u (approved):

AaronFeickert (approved):

AaronFeickert:
I like the updated approach better, in part because it makes it more clear by the naming of the binaries and environment variable whether the user is running the safe or reduced parameter set.

PR #172, opened 2026-04-16
Feat/setup progress log

sapinb:
## Description
Improve evaluator resilience and job progress observability

- [x] Add progress tracking to RPC status: Expand TablesetStatus details to show per-step progress (e.g. "GeneratingShares (3/5)") for both garbler and evaluator, instead of just the step name.
- [x] Add RetryEvalState wrapper: Introduce a wrapper around EvaluatorState that auto-retries FDB reads on transaction timeout errors, reducing transient failures during garbling table evaluation.
- [x] Add failure logging: Log error details on handle_receive_garbling_table failure paths for easier debugging.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency update
- [ ] Security fix

## Notes to Reviewers
changes in `fix: create new txn after receiving tables in handle_receive_garbling_table` commit are superseded by `RetryEvalState` changes.

## Checklist


Zk2u:
I pushed `fa20b1c` to remove `RetryEvalState` from this branch.

Why `RetryEvalState` is not acceptable:
- it relied on unconditional `unsafe impl Send/Sync` over `&SP` + `UnsafeCell<SP::EvaluatorState>`, which is not a sound way to express FDB retry semantics
- it tried to treat evaluator state handles as resumable/retryable across long-running jobs, but FDB `on_error()` semantics require a fresh transaction and a full rerun of the transaction body
- many job executors are not single retryable transaction bodies; they do long I/O and other side effects between state accesses, so a generic transaction wrapper is the wrong abstraction here

Plan from here:
1. Jobs should use the scheduler’s built-in retry path for transient failures instead of an unsafe generic state wrapper.
2. For long-running evaluator jobs that need fresh state after long I/O, we should reacquire state explicitly at the specific checkpoint where that is needed rather than pretending one transaction can be resumed.
3. Separately, we should investigate proper FoundationDB `on_error()` integration in `sm-executor`, where the unit of work is much closer to a real retryable transaction body (load state, apply STF, commit).

I’m opening follow-up issues for the job-side cleanup and the `sm-executor` FDB retry investigation so this PR can stay focused on setup visibility/progress.

sapinb:
@Zk2u RetryEvalState is only meant to be used in the jobs, which only read data from fdb and does not care about transaction semantics anyway. I strongly argue that we have some form of auto retry mechanism here. Otherwise this job run is currently guaranteed to fail. Please use another ticket/PR to add a better fix, but let us merge a working fix with this PR in the meantime

Zk2u:
It was planned from the start to use per call transactions for the restore and job system, but was never implemented. This is sorted in #179. I've dropped RetryEvalState from this PR and will merge it in without. 179 sorts the rest out + more.

Zk2u (approved):

PR #179, opened 2026-04-18
Migrate to FASM 0.4 and split read-only state paths

### LINKED ISSUE #176: job executors: use scheduler retry / explicit state reacquisition instead of generic FDB retry wrappers
## Summary

The `RetryEvalState` experiment on `#172` is the wrong abstraction for job executor retry behavior.

Job executors already have a built-in retry mechanism via `HandlerOutcome::Retry`. A generic wrapper that tries to make evaluator state reads transparently retryable across long-running jobs is both misleading and unsafe.

## Why

FoundationDB `Transaction::on_error()` is not a "resume this transaction" primitive. It means:

1. the current transaction attempt failed with a retryable error
2. apply FDB's retry/backoff policy
3. create a fresh transaction
4. rerun the entire transaction body from the start

Many job executors are not single retryable transaction bodies. They do long-running work in between state accesses, including:
- network sends/receives
- bulk table transfer
- table-store writes
- circuit work

For those jobs, a late FDB error cannot be safely handled by pretending the state handle is resumable.

## Desired direction

1. Do not use a generic retry wrapper like `RetryEvalState`.
2. Let job executors continue using the scheduler's existing retry path for transient failures.
3. In the few long-running paths that need fresh state after long I/O, reacquire a fresh state handle explicitly at the exact checkpoint where it is needed.
4. Audit evaluator/garbler executors for places where explicit state reacquisition would be preferable to whole-job retry.

## Acceptance criteria

- no generic `RetryEvalState`-style wrapper exists in job executors
- job retry remains driven by the scheduler / `HandlerOutcome::Retry`
- any state refresh in a job executor is explicit and local to the relevant long-I/O boundary
- no `unsafe Send/Sync` wrapper is used to paper over transaction retry semantics

### LINKED ISSUE #177: sm-executor: use FoundationDB transact-style retry for STF apply/commit units
## Summary

`sm-executor` should use FoundationDB transaction retry semantics for STF application / commit paths.

This should now be framed around **`Database::transact(...)`-style full-transaction retry**, not a vague `on_error()` patch around individual operations.

We only use `on_error()` today in FDB directory/keyspace initialization loops. Normal mutable state application in `sm-executor` still goes through:
- acquire mutable state handle
- run STF
- plain `tx.commit().await`
- then submit follow-up work

## Why this matters

The `sm-executor` unit of work is a strong fit for FDB transaction retry semantics:

1. load mutable state
2. apply deterministic STF transition
3. commit state
4. only after successful commit, emit external side effects (job submissions / acks)

That is exactly the kind of operation that can be retried safely as a whole transaction body.

## Updated reasoning

This issue is distinct from the job/restore read model.

Jobs and restore are read-only and should move toward fresh-read state implementations. `sm-executor` is different: it is the transactional state-application engine.

Because FASM is explicitly designed so that STF has no external side effects and only emits action descriptions, rerunning STF inside a transaction retry loop is semantically acceptable **as long as actions are only submitted after the transaction succeeds**.

## Scope

Investigate and implement a proper FoundationDB retry model for `sm-executor` STF apply/commit units.

That means:
- retry the **entire** state-application unit
- not just `commit()`
- not just one failed storage call
- and not by using ad hoc job-style retry wrappers

The likely target is a `Database::transact(...)`-style integration, or an equivalent storage abstraction that provides the same semantics.

## Questions to answer

- Which `sm-executor` paths can be wrapped as one retryable transaction body?
- Are all external side effects already sequenced strictly after commit?
- What storage-provider abstraction changes are needed so this works cleanly for FDB without infecting non-FDB backends?
- Can event application, completion application, and inbound-request application all use the same transactional wrapper?

## Acceptance criteria

- `sm-executor` retries **full STF apply/commit units**, not partial operations
- external side effects are not duplicated across retry
- FDB retry semantics are handled in the right layer instead of through ad hoc wrappers
- the design remains coherent for non-FDB backends
- the resulting implementation is clearly distinct from the read-only job/restore state model

### LINKED ISSUE #178: FDB read-only StateRead impls: fresh transaction per method call for jobs and restore
## Summary

For the FoundationDB backend, read-only `StateRead` implementations should use a **fresh transaction per method call**.

This should be used by:
- the **job system**
- the **SM restore path**

This was the original intended design for the job system. We had discussed and planned for jobs to use fresh read-only transactions because jobs only read SM state, but we missed implementing it that way and instead ended up with long-lived state handles.

Now that FASM 0.4 supports a separate `RestoreState`, the same read-only model should also be used for restore.

## Problem

Today, many job paths do this:

1. call `storage.evaluator_state(peer_id)` or `storage.garbler_state(peer_id)` once
2. reuse that handle across many `get_*` calls
3. sometimes do long-running work in between those reads

For FoundationDB-backed storage, that means a single underlying transaction can stay alive across:
- many sequential reads
- network waits
- table-store / S3 operations
- circuit setup / interpolation work

That is a poor fit for FoundationDB and leads to transaction timeout / expiry problems.

The attempted `RetryEvalState` approach in `#172` was the wrong abstraction:
- unsafe `Send/Sync`
- generic retry wrapper over the wrong semantic boundary
- treated state handles as if they were resumable
- hid the real read model instead of making it explicit

## Why this is safe

For the current protocol, using fresh transactions per read method call is safe for jobs because:
- jobs only **read** SM state
- the data consumed by jobs is stable enough by the time the job is emitted
- mixed snapshots cause retry/unavailability, not silent corruption
- the scheduler already has a built-in retry mechanism for jobs

Restore has the same fundamental property:
- restore is read-only over SM state
- it reconstructs actions from committed state
- it does not mutate SM state

So jobs and restore should share the same read-only FDB state model.

## Proposed design

Do **not** add new traits.

Keep the existing garbler/evaluator `StateRead` traits, but change the FDB-backed read-only implementations so that each read method:

1. acquires a fresh underlying FDB transaction
2. performs exactly one read operation
3. drops the transaction

This change belongs in the **FDB storage crate**, not just in the job executor layer, because it now serves two consumers:
- jobs
- restore

## Notes

- This issue is about the **read-only** state model.
- STF/event/completion application in `sm-executor` remains a separate concern and is tracked separately.
- `stream_all_deposits()` also needs a read-only long-running-safe design; it cannot just be a giant long-lived FDB transaction stream.

## Scope

1. Implement FDB read-only `StateRead` handles with fresh transaction per method call.
2. Ensure the design also supports restore, not just jobs.
3. Switch job executors to use these read-only semantics by default.
4. Switch restore to use read-only state once the FASM 0.4 `RestoreState` migration is in place.
5. Remove retry-wrapper approaches such as `RetryEvalState`.

## Acceptance criteria

- read-only FDB `StateRead` handles do not hold one transaction open across the whole read phase of a job or restore pass
- jobs use fresh transactions per read method call
- restore uses the same read-only FDB state model
- existing `StateRead` trait surface is preserved
- no unsafe retry wrapper remains in the job system or restore path
- long-running job/restore reads no longer depend on one long-lived transaction surviving across network / S3 / circuit work

Zk2u:
## Summary
- migrate Mosaic to `fasm` 0.4 and declare separate `RestoreState` types for the protocol state machines
- add read-only FDB state handles that open a fresh transaction per read call and use them for job execution and SM restore
- make `sm-executor` retry whole STF units on retryable storage failures instead of using ad hoc stale-transaction retry wrappers

Closes #176
Closes #177
Closes #178

## Commit stack
1. `build: migrate to fasm 0.4`
2. `storage/fdb: add read-only state handles`
3. `job/executors: remove stale transaction retries`
4. `sm-executor: restore from read-only state`
5. `sm-executor: retry stf units on retryable storage errors`

## Design
- jobs and restore now use read-only state access with fresh FDB transactions per method call
- STF apply remains on mutable state handles, but retries the whole logical STF unit when storage marks the failure as retryable
- `Commit` is preserved with the stronger contract that success means durable commit and error means no partial visible state

## Validation
- `cargo check -p mosaic-sm-executor -p mosaic-cac-protocol -p mosaic-cac-types -p mosaic-storage-api -p mosaic-storage-inmemory -p mosaic-storage-fdb -p mosaic-storage-kvstore`
- `cargo test -p mosaic-sm-executor --lib`
- `cargo test -p mosaic-storage-fdb`
- `cargo clippy -p mosaic-sm-executor -p mosaic-storage-fdb -p mosaic-storage-kvstore --tests --all-targets -- -D warnings`
- `cargo clippy -p mosaic-cac-protocol -p mosaic-cac-types -p mosaic-storage-api -p mosaic-storage-inmemory --tests --all-targets -- -D warnings`

Note: `cargo test -p mosaic-cac-protocol` reported all 24 unit tests green, but the process lingered on teardown in this environment after success output.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2620,6 +2620,7 @@ version = "0.1.0"
 dependencies = [
  "async-trait",
  "foundationdb",
+ "mosaic-cac-types",
  "mosaic-net-svc-api",
  "mosaic-storage-api",
  "mosaic-storage-kvstore",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -373,7 +373,7 @@ dependencies = [
  "bitflags 2.10.0",
  "cexpr",
  "clang-sys",
- "itertools 0.10.5",
+ "itertools 0.13.0",
  "log",
  "prettyplease",
  "proc-macro2",
@@ -1164,9 +1164,9 @@ dependencies = [
 
 [[package]]
 name = "fasm"
-version = "0.3.1"
+version = "0.4.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "4eeeb720348b3e0675242d3c09fdb75be4bc216ff01381348d22c4fffee571a2"
+checksum = "ec91f05d045b6a9aa3964796b78f046e056779af00c7d03041e8b2abbd4da89f"
 
 [[package]]
 name = "fastbloom"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -104,7 +104,7 @@ bytes = "1.10"
 ckt-fmtv5-types = { git = "https://github.com/alpenlabs/ckt" }
 ckt-gobble = { git = "https://github.com/alpenlabs/ckt" }
 criterion = "0.5"
-fasm = "0.3.0"
+fasm = "0.4.0"
 # TODO(alpenlabs): switch back to crates.io foundationdb >= 0.11 once released.
 
 foundationdb = { git = "https://github.com/foundationdb-rs/foundationdb-rs.git", rev = "ed274ba533e15ec02e4ab8dd4defca7d18eaf355", features = [
--- a/crates/cac/protocol/Cargo.toml
+++ b/crates/cac/protocol/Cargo.toml
@@ -25,9 +25,9 @@ tracing.workspace = true
 
 [dev-dependencies]
 mosaic-cac-types = { workspace = true, features = ["test-utils"] }
+mosaic-storage-api.workspace = true
 mosaic-storage-inmemory.workspace = true
 mosaic-job-executors.workspace = true
-mosaic-storage-api.workspace = true
 mosaic-job-api.workspace = true
 mosaic-net-svc-api.workspace = true
 mosaic-net-svc.workspace = true
--- a/crates/cac/protocol/src/error.rs
+++ b/crates/cac/protocol/src/error.rs
@@ -1,6 +1,6 @@
 use std::error::Error;
 
-use mosaic_cac_types::DepositId;
+use mosaic_cac_types::{DepositId, RetryableStorageError};
 
 /// State machine error
 #[derive(Debug, thiserror::Error)]
@@ -27,8 +27,15 @@ pub enum SMError {
     #[error("CRITICAL: State is inconsistent with expectations: {0}")]
     StateInconsistency(String),
     /// Error while accessing storage.
-    #[error("Error while accessing storage: {0}")]
-    Storage(Box<dyn Error>),
+    #[error("Error while accessing storage: {source}")]
+    Storage {
+        /// Whether the underlying storage failure is safe to retry by
+        /// discarding the current STF attempt and rerunning it from the start.
+        retryable: bool,
+        /// Original storage error.
+        #[source]
+        source: Box<dyn Error + Send + Sync>,
+    },
 }
 
 impl SMError {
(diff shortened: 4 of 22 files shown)


sapinb (approved):
Looks great.

Only concern if stf restore now running outside a transaction can be problematic. 
Also, this cuts off the option of rerunning restore outside startup.

PR #180, opened 2026-04-22
mosaic: fail closed on critical thread exit

### LINKED ISSUE #181: Fail closed when critical Mosaic threads exit unexpectedly
## Problem

Mosaic currently keeps running if a critical background thread dies unexpectedly.

In practice this means a panic or fatal thread exit can show up in logs, but the process stays alive even though it can no longer function correctly. This leaves the node in a degraded state instead of failing closed.

Observed examples include:
- `sm-executor` thread death
- job-scheduler thread death
- internal scheduler worker/coordinator thread death that effectively disables job execution
- RPC or net-service thread death

## Desired behavior

If any critical long-lived thread exits unexpectedly, Mosaic should:

1. surface a clear fatal log
2. initiate shutdown of the remaining components
3. exit the process with an error

This should apply not only to top-level component threads, but also to internal thread dependencies whose death makes the owning component non-functional.

## Scope

Initial scope:
- propagate fatal worker/coordinator thread exits up to the owning component
- expose component liveness at the controller level
- make `bin/mosaic` wait on either:
  - shutdown signal, or
  - unexpected critical component exit
- on unexpected exit, perform best-effort shutdown of the remaining components and return an error

## Notes

This is similar to the fail-fast supervision pattern already used in other Alpen/Bridge services.

The intent is operational correctness, not graceful recovery: if a critical thread dies, Mosaic should fail closed rather than continue running in a broken state.

Zk2u:
## Summary

Fail closed when a critical Mosaic thread exits unexpectedly instead of leaving the process running in a degraded state.

Closes #181.

## What changed

- propagate scheduler pool worker panics as fatal scheduler faults
- propagate garbling worker and garbling coordinator panics as fatal scheduler faults
- expose controller liveness checks for `rpc`, `sm-executor`, and `job-scheduler`
- make `bin/mosaic` wait on either shutdown signal or unexpected critical component exit
- on unexpected exit, perform best-effort shutdown of the remaining components and return an error

## Behavior

After this change:

- internal scheduler worker/coordinator thread death bubbles up to the scheduler
- unexpected exit of `rpc`, `net-svc`, `sm-executor`, or `job-scheduler` causes Mosaic to shut down and exit nonzero
- one dead component no longer prevents best-effort shutdown of the others

## Validation

- `cargo check -p mosaic -p mosaic-job-scheduler -p mosaic-sm-executor -p mosaic-net-svc`
- `cargo clippy -p mosaic -p mosaic-job-scheduler -p mosaic-sm-executor -p mosaic-net-svc --all-targets -- -D warnings`

### DIFF
--- a/bin/mosaic/src/main.rs
+++ b/bin/mosaic/src/main.rs
@@ -6,7 +6,11 @@ mod rpc;
 use std::{
     env,
     path::PathBuf,
-    sync::{Arc, mpsc},
+    sync::{
+        Arc,
+        mpsc::{self, RecvTimeoutError},
+    },
+    time::Duration,
 };
 
 use anyhow::{Context, Result, bail};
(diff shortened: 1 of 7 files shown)


sapinb (approved):
ACK c257c3a157957db83e323c0391d61bb28ac07f33

Issue #190, opened 2026-04-22
Worker permits should not be leaked by permanently stalled jobs


Zk2u:
## Summary

Pool concurrency is enforced by explicit permit return at task tail. Permanently stalled tasks never hit that return path, so enough stuck tasks can wedge the worker.

This is the reviewed outcome for audit finding `#48181`.

## Impact

- Long-lived capacity degradation or exhaustion in light/heavy worker pools.
- Under our operating model, this is a medium-severity availability issue.

## Proposed fix

Represent permits with a drop guard so they are automatically returned on all normal unwind paths.

Also add task-level deadlines/cancellation for worker jobs so permanently stalled tasks cannot hold permits forever.

Note:
- the existing fail-closed thread-supervision work helps with panic cases
- it does not solve ordinary indefinite stall cases

## Acceptance criteria

- No task can leak a worker permit indefinitely through ordinary stall behavior.

AaronFeickert:
@Zk2u is this still relevant?

PR #197, opened 2026-04-22
net/client: bound worker-side send stalls

### LINKED ISSUE #183: Outbound worker sends lose open and ACK deadlines on monoio workers
## Summary

`NetClient::send()` only applies stream-open and ACK deadlines when a Tokio runtime is present. Job-scheduler workers run under monoio, so those deadlines are effectively disabled there. A misbehaving peer can therefore pin worker slots indefinitely by withholding ACKs.

This is the reviewed outcome for audit finding `#48100`.

## Impact

- Light-pool capacity can be exhausted by hung outbound protocol sends.
- This requires an authenticated/misbehaving peer, not a public unauthenticated caller.
- Under our operating model, this is a medium-severity availability issue.

## Proposed fix

Make send/open/ack deadlines runtime-agnostic.

Implementation options:
- replace the current Tokio-only timeout helper with a timeout mechanism that works under both Tokio and monoio
- or push deadlines down into the transport/client abstraction so callers cannot silently lose them based on runtime choice

At minimum, all job-system outbound sends should enforce:
- stream open timeout
- ACK timeout

## Acceptance criteria

- On monoio workers, stalled open/ACK waits time out and return control to the worker so permits are released and the job can retry/fail normally.

Zk2u:
## Summary
- make `NetClient::send()` enforce open and ack deadlines on monoio workers as well as Tokio callers
- add a permit guard so worker permits are released on panic and early exit
- add a monoio regression test proving sender-side ack timeouts now fire on worker runtimes

## Issues
Closes #183.
Refs #190.

## Validation
- `cargo test -p mosaic-net-client -p mosaic-job-scheduler`
- `cargo check -p mosaic -p mosaic-net-client -p mosaic-job-scheduler`
- `cargo clippy -p mosaic-net-client -p mosaic-job-scheduler -p mosaic --tests --all-targets -- -D warnings`

## Notes
- This PR intentionally does not add a generic worker execution timeout.
- `ReceiveGarblingTable` stall handling remains in the separate bulk-transfer hardening work.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -99,3 +99,6 @@ fuzz-*.log
 # Docker volume data
 docker/vol/fdb/data/
 docker/vol/mosaic-*/tables/
+
+# local scratch docs
+tmp/
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1376,6 +1376,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "037711b3d59c33004d3856fbdc83b99d4ff37a24768fa1be9ce3538a1cde4393"
 
 [[package]]
+name = "futures-timer"
+version = "3.0.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f288b0a4f20f9a56b5d1da57e2227c661b7b16168e2f72365f57b63326e29b24"
+
+[[package]]
 name = "futures-util"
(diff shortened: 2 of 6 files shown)


sapinb (approved):
ACK a32fd070e2f18ac9ac59b9b82fcca428efcf8c93

PR #201, opened 2026-04-29
functional-tests: support MOSAIC_PATH env var for mosaic binary path

### LINKED ISSUE #200: Functional tests require Mosaic binary to be in path
Functional tests currently require that the Mosaic binary be in path. This is automatically handled in CI, but for local testing needs to be done manually.

It would be useful to support arbitrary binary paths via an environment variable, similarly to how FDB is handled.

mvanhorn:
Resolves #200.

Functional tests previously hard-coded `"mosaic"` as the binary invoked by the Mosaic factory. CI handles this by adding the binary to PATH, but local runs needed the manual workaround of copying the binary into PATH or adjusting it.

## Change

Add `MOSAIC_PATH = os.environ.get("MOSAIC_PATH", "mosaic")` at module load in `functional-tests/factories/mosaic/__init__.py`, mirroring the existing `FDBSERVER_PATH` / `FDBCLI_PATH` pattern in `functional-tests/factories/fdb.py`. Use the variable in the spawn `cmd` list.

Default behaviour is unchanged when the env var is unset (resolves to `"mosaic"` on PATH).

## Local usage

```bash
export MOSAIC_PATH=$PWD/target/debug/mosaic
pytest functional-tests/
```

Diff: 7 lines.

### DIFF
--- a/functional-tests/factories/mosaic/__init__.py
+++ b/functional-tests/factories/mosaic/__init__.py
@@ -11,6 +11,11 @@ from services import ProcServiceWithEnv, get_fdb_env
 
 from .mosaic_config import *
 
+# Environment variable to override the mosaic binary path.
+# Mirrors the FDBSERVER_PATH / FDBCLI_PATH pattern in factories/fdb.py.
+# Useful when the binary is not on PATH (e.g., a debug build under target/debug).
+MOSAIC_PATH = os.environ.get("MOSAIC_PATH", "mosaic")
+
 
 @dataclass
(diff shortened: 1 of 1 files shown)


AaronFeickert (approved):

AaronFeickert:
Sounds like @Zk2u would like to make a change to the default path before we merge this.

Zk2u (changes requested):
default should be changed to `target/release/mosaic` before merge

Zk2u (approved):

Issue #220, opened 2026-05-08
Per-peer protocol-message rate limiting is not enforced beyond connection establishment


Zk2u:
## Summary

Mosaic's network layer rate-limits at connection establishment (per-IP `InboundRejectTracker` + `MAX_CONCURRENT_INCOMING_HANDSHAKES = 128` semaphore + per-connection `MAX_CONCURRENT_BIDI_STREAMS = 100` QUIC transport limit), but enforces no rate limit on the protocol-message stream that a connected peer can drive.

Once authenticated, a peer can open up to 100 concurrent bidi streams and submit valid-looking protocol messages (`TableTransferRequestMsg`, `ChallengeMsg`, `ChallengeResponseMsgChunk`, etc.) as fast as the wire allows. Each accepted message can fan out into expensive internal work — a single `TableTransferRequestMsg`, for example, dispatches a full `TransferGarblingTable` action that opens a bulk stream and processes the entire circuit.

## Why this matters now

We have decided to make internal job-system channels (scheduler/worker completion + submission) effectively unbounded, since the realistic worst-case sizing is far below memory pressure (~100B per entry, low thousands of entries even at large operator counts). That decision is only safe if **upstream** admission control is real — i.e. the volume of work entering the system per peer is bounded somewhere. Today, that bound only exists at the connection layer; the protocol layer trusts authenticated peers to be well-behaved.

This is the upstream complement to the bounded-queue deadlock work tracked in #182 and the new peer-offline deadlock issue: tighten the boundary so the internal removal of backpressure stays safe.

## Impact

- A misbehaving or compromised configured peer can flood the protocol layer with valid-looking requests, driving sustained job-system load and saturating workers / bulk-transfer slots / table-store I/O.
- This is reachable by any of our configured external operators if their key is compromised, and bilateral if our own keys are compromised.
- Severity: high. Reaches the same surface as the protocol-soundness and resource-exhaustion findings, but more general.

## Proposed fix

Per-peer rate limiting on accepted protocol messages, applied between stream-payload acceptance and protocol dispatch.

Reasonable implementation options:

- token-bucket per peer in net-svc, before routing payloads to the protocol stream queue
- separate buckets per message kind (or at least separate buckets for cheap header messages vs. expensive fan-out messages like `TableTransferRequestMsg` and deposit-init)
[...]

PR #198, opened 2026-04-23
job/executors: harden garbling table receive path

Zk2u:
Closes #188.
Closes #193.

## Summary
- add runtime-agnostic bulk open/read timeout helpers in `mosaic-net-client`
- add explicit expectation cancellation on bulk-open timeout
- add slow-path warning logs and hard timeouts in `ReceiveGarblingTable`
- enforce exact ciphertext length using the circuit header `and_gates` count
- reset timed-out bulk streams and delete partial table state on failure

## Validation
- cargo test -p mosaic-net-client --tests
- cargo test -p mosaic-job-executors --lib
- cargo check -p mosaic -p mosaic-job-executors -p mosaic-net-client -p mosaic-net-svc-api
- cargo clippy -p mosaic-job-executors -p mosaic-net-client -p mosaic-net-svc-api --tests --all-targets -- -D warnings
- cargo +nightly fmt --all --check

PR #222, opened 2026-05-08
job/scheduler,sm-executor,net/svc: unbound internal channels

### LINKED ISSUE #182: Scheduler dispatch can be head-of-line blocked by oversized garbling batches
## Summary

The global job scheduler dispatch loop awaits garbling-coordinator submission inline for each circuit action. Because the garbling coordinator submission channel is bounded and only drains between passes, one large but valid garbling batch can head-of-line block new dispatch for unrelated work.

This is the reviewed outcome for audit finding `#48090`.

## Impact

- New light/heavy jobs for other peers can be delayed behind garbling pass backpressure.
- This blocks new dispatch, not work already running in pools/coordinator.
- Under our operating model, this is best treated as a medium-severity scheduler fairness/liveness bug.

## Proposed fix

Decouple global batch dispatch from garbling pass backpressure.

Reasonable implementation options:
- enqueue garbling jobs to an internal scheduler-side buffer without awaiting each coordinator submission inline
- or split garbling dispatch onto a dedicated forwarding task so the main scheduler loop can keep draining later batches
- keep a bounded queue and explicit overflow/backpressure policy for garbling submissions instead of reusing the coordinator channel directly from the main dispatch loop

## Acceptance criteria

- A valid garbling batch larger than `max_concurrent * 2` does not block unrelated later light/heavy batch dispatch for the duration of a garbling pass.

### LINKED ISSUE #221: Peer-offline failure bursts can deadlock the SM executor through bounded job-system queues
## Summary

A peer going offline mid-session can produce a burst of job failures large enough to deadlock the SM executor against the job scheduler through the system's bounded inter-component queues. The issue affects all peers, not just the offline one — including peers with no protocol dependency on the offline peer.

Hit in production on 2026-05-06 during a coordinated restart with an external operator: ~165 in-flight `SendCommitMsgChunk` jobs for the offline peer failed near-simultaneously, overlapping with another peer's ~165 polynomial-commitment completions. The system reached a state where S3 stayed empty for 30+ minutes until the node was restarted.

## Mechanism

Three bounded `kanal` channels form a cycle that wedges under bursty failure load:

- `completion_queue_size` (default 256, `crates/job/scheduler/src/scheduler.rs:140`) — workers → SM executor
- `submission_queue_size` (default 256, `crates/job/scheduler/src/scheduler.rs:137`) — SM executor → scheduler dispatch
- The scheduler dispatch loop can additionally be back-pressured by the garbling-coordinator submission channel (#182).

The SM executor processes all event sources (job completions, inbound network requests, operator commands, retries) through a single `monoio::select!` loop in `crates/sm-executor/src/lib.rs:302-403`, one event at a time. Sequence:

1. Failure burst: many in-flight jobs for the offline peer fail near-simultaneously. Workers push N completions into `completion_tx` (cap 256). With concurrent traffic from healthy peers, the channel fills.
2. Each completion the SM executor handles runs the STF, which may call `job_handle.submit()` → `submit_tx` (cap 256). If the scheduler dispatch loop is itself backpressured (e.g. by #182), `submit_tx.send().await` blocks the executor mid-completion.
3. Executor blocked → `completion_tx.send().await` blocks workers → completions cannot drain. Workers blocked → `submit_tx` consumer makes no progress.
4. The executor's inbound-network branch never fires because the loop is parked inside the completion branch's `submit().await`. Inbound `TableTransferRequestMsg` from any peer (including healthy internal peers) is not dropped at the channel level — it simply never gets selected. Externally indistinguishable from a drop.

Independent reproduction: with `submission_queue_size` and `completion_queue_size` reduced to 16, setup gets stuck at `GeneratingTableCommitments (17/181)` — exactly `channel_size + 1`. With size 8, stuck at 9/181. Classic bounded-queue deadlock signature.

## Impact

- A single peer dropping during active garbling can stall the entire node, including sessions between healthy internal peers that have zero protocol dependency on the offline peer.
- The stall persists until manual restart.
- Reachable by any normal operational event in any external-operator integration. Bilateral.
- Severity: high (confirmed production outage).

## Bandaid currently deployed

Internal stack has been bumped to:

- `submission_queue_size = 1024`
- `completion_queue_size = 1024`
- `command_queue_size = 1024`

Widens the deadlock window proportionally to peer count but does not eliminate the cycle. Acceptable as a workaround; not a fix.

## Proposed fix

Make the internal job-system channels unbounded and rely on boundary admission control (peer protocol-message rate limiting, **#220**) to bound the actual fan-in.

Channels to flip to `kanal::unbounded_async`:

- `JobScheduler::submit_tx` (SM executor → scheduler)
- `JobScheduler::completion_tx` (workers → SM executor)
- `JobScheduler::fault_tx` (workers → scheduler fault)
- `GarblingCoordinator::submit_tx` (scheduler → garbling coord) — also closes #182
- `SmExecutor::command_tx` (operator/RPC → SM executor)
- `NetService::command_tx` (handles → net-svc service)

External-facing channels stay bounded:

- `NetService::protocol_stream_tx` and per-stream `payload_tx` / `buf_return_tx` — peer-fed
- Per-peer rate limiting handled separately in #220

Add `.len()` gauges on every channel flipped to unbounded so genuine pathological growth is visible before OOM.

## Acceptance criteria

- A peer dropping mid-garbling does not stall sessions between healthy peers.
- The deadlock signature ("stuck at exactly channel_size + 1") cannot be reproduced under failure bursts at production-realistic queue sizes.
- Channel-length metrics are exported for the now-unbounded internal channels.

## References

- 2026-05-07 production incident
- Related: #182 (scheduler-side bounded-queue HOL — closed by same fix), #183 (monoio worker timeouts — failure-burst rate amplifier), #184 (inverse: unbounded worker queues — to be reframed via #220), #220 (per-peer protocol-message rate limiting — boundary admission)

Zk2u:
## Summary

Bounded internal channels with `send().await` create cycle-deadlock risk between trusted in-process components. The 2026-05-07 production incident (#221) and the scheduler-dispatch HOL block in #182 are both instances of the same pattern: a producer that submits inline while consuming from another bounded channel can wedge the system under bursty load.

This PR flips the internal job-system and control-plane channels to unbounded:

- `JobScheduler::submit_tx` — SM executor → scheduler dispatch
- `JobScheduler::completion_tx` — workers → SM executor
- `JobScheduler::fault_tx` — workers → scheduler fault
- `GarblingCoordinator::submit_tx` — scheduler dispatch → garbling coord (also closes #182)
- `SmExecutor::command_tx` — operator/RPC → SM executor
- `NetService::command_tx` — handles → net-svc service

External-facing channels stay bounded: `NetService::protocol_stream_tx` and the per-stream `payload_tx` / `buf_return_tx` are peer-fed and rely on the upcoming per-peer protocol-message rate limit (#220) for boundary admission control.

`submission_queue_size`, `completion_queue_size`, and `command_queue_size` config fields are retained for compatibility but are no longer load-bearing. `.len()` instrumentation on the now-unbounded channels is left as follow-up.

## Issues
Closes #182.
Closes #221.

## Validation
- `cargo check --workspace`
- `cargo test -p mosaic-job-scheduler -p mosaic-sm-executor -p mosaic-net-svc -p mosaic-net-client -p mosaic-job-executors`

## Notes
- Functional tests not run locally; expecting CI to confirm the deadlock signature is gone (#198's functional-test failures should also clear once this lands and #198 rebases on top).
- Per-peer rate limiting (#220) is the upstream complement to this change and is filed but not started.

### DIFF
--- a/crates/job/scheduler/src/garbling/mod.rs
+++ b/crates/job/scheduler/src/garbling/mod.rs
@@ -300,7 +300,12 @@ impl GarblingCoordinator {
         completion_tx: kanal::AsyncSender<JobCompletion>,
         fault_tx: kanal::AsyncSender<SchedulerFault>,
     ) -> Self {
-        let (submit_tx, submit_rx) = kanal::bounded_async(config.max_concurrent * 2);
+        // Unbounded: the scheduler dispatch loop forwards each circuit
+        // action through this channel inline. Bounding it head-of-line
+        // blocks unrelated dispatch when a garbling pass is in progress
+        // (see #182). Volume is bounded upstream by per-peer protocol-message
+        // rate limiting (#220).
+        let (submit_tx, submit_rx) = kanal::unbounded_async();
 
         let coordinator_thread_name = "garbling-coordinator".to_string();
         let thread = std::thread::Builder::new()
--- a/crates/job/scheduler/src/scheduler.rs
+++ b/crates/job/scheduler/src/scheduler.rs
@@ -35,8 +35,15 @@ pub struct JobSchedulerConfig {
     /// Garbling coordinator: coordinated topology reads + garbling.
     pub garbling: GarblingConfig,
     /// Capacity of the submission channel.
+    ///
+    /// No longer load-bearing: the submission channel is now unbounded to
+    /// avoid bounded-queue cycle deadlocks with the SM executor (see #221).
+    /// Retained for config compatibility; ignored at runtime.
     pub submission_queue_size: usize,
     /// Capacity of the completion channel.
+    ///
+    /// No longer load-bearing: the completion channel is now unbounded for
+    /// the same reason as `submission_queue_size`.
     pub completion_queue_size: usize,
 }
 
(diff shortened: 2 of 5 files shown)


sapinb (approved):
ACK 306106bf62e8293462e87da8868ece9d53fd85d4

PR #224, opened 2026-05-08
cac/types: redact Seed in Debug and Display formatting

### LINKED ISSUE #213: Job retry/drop log path leaks raw garbling seeds via Debug formatting
## Summary

The job-completion retry/drop path in `crates/sm-executor/src/lib.rs` logs `action_id = %completion.action_id()` on both requeue and drop. `PendingJobCompletion::action_id()` formats the underlying action ID with `Debug`. Garbler action IDs include `Seed` and `GarblingSeed` (which is just an alias for `Seed`), and `Seed`'s `Debug` impl prints the complete raw value as hex.

Any storage/commit failure that triggers completion requeueing, or any STF failure that triggers dropping, therefore writes secret setup/garbling seeds to warning-level logs.

This is the reviewed outcome for Codex security finding [`e7cc453e`](https://chatgpt.com/codex/cloud/security/findings/e7cc453e8270819195d44ea10ae0e4fb).

## Impact

- Garbling/setup seeds are protocol-secret material. Their exposure to log aggregation, support bundles, or anyone with operator log access weakens cut-and-choose and downstream signing security.
- Triggered by ordinary error paths (storage commit failures, STF failures), not by adversarial action.
- Severity: high (secret material in logs).

## Proposed fix

Two layers:

1. Make `Seed` impossible to leak by accident:
   - replace `Seed`'s `Debug` impl with a redacted form (e.g. `Seed(<redacted>)` or first-N-bytes-of-hash)
   - add an explicit `reveal()` or `expose_secret()` accessor for the very few legitimate consumers
   - consider wrapping the inner bytes with a `secrecy::SecretVec`-style type to enforce the discipline

2. Audit the SM-executor log call sites that include action IDs:
   - the retry/drop paths in `sm-executor/src/lib.rs`
   - any other `?action` / `%action_id` formatting elsewhere
   - prefer logging a stable non-secret identifier (action kind + index, or a hash of the action ID) instead of the action ID directly

## Acceptance criteria

- Default `Debug`/`Display` of `Seed` and `GarblingSeed` does not print the raw secret bytes.
- The job retry/drop log paths do not emit anything that contains a raw seed.
- Existing logs that need an action identifier emit a non-secret form (kind + index, or hash).

Zk2u:
## Summary

Closes #213.

`Seed` carries protocol-secret material (garbling / polynomial seeds). Its previous `Debug` / `Display` impls printed the raw bytes as hex, which meant any log line that formatted an enclosing type via `{:?}` / `{}` could leak the seed. The SM-executor's job-completion retry/drop log paths in particular format `ActionId` (which contains `GarblingSeed`) on every storage-commit / STF failure, writing seeds to WARN-level logs.

## Changes

- `Seed::Debug` and `Seed::Display` now both render `Seed(<redacted>)`.
- Three regression tests assert neither formatter ever emits hex.
- The explicit accessors (`to_hex`, `as_bytes`, `to_bytes`) are unchanged — these remain the only sanctioned way to reveal seed material.

`GarblingSeed` is a type alias for `Seed`, so this fix covers both surfaces.

## Issues
Closes #213.

## Validation
- `cargo check --workspace`
- `cargo test -p mosaic-cac-types` (40 lib tests pass; 3 new redaction tests included)

### DIFF
--- a/crates/cac/types/src/seed.rs
+++ b/crates/cac/types/src/seed.rs
@@ -84,17 +84,58 @@ impl TryFrom<&[u8]> for Seed {
 }
 
 // ---------------------------------------------------------------------------
-// Display / Debug
+// Display / Debug — both deliberately redact secret bytes
 // ---------------------------------------------------------------------------
+//
+// `Seed` carries protocol-secret material (garbling seeds, polynomial seeds).
+// Default formatting (`{}`, `{:?}`, derived `Debug` on enclosing types) must
+// not leak these bytes into logs, panic messages, or operator-visible
+// diagnostics. Callers that need the raw bytes must reach for the explicit
+// accessors (`to_hex`, `as_bytes`, `to_bytes`).
 
 impl fmt::Display for Seed {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        f.write_str(&hex::encode(self.0))
+        f.write_str("Seed(<redacted>)")
     }
 }
 
 impl fmt::Debug for Seed {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        write!(f, "Seed({})", hex::encode(self.0))
+        f.write_str("Seed(<redacted>)")
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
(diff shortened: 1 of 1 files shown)


sapinb (approved):
ACK 62a02b441a4ef15d848b965cab859d12f16e8850

PR #225, opened 2026-05-08
docker: bind FoundationDB to loopback only in Compose

### LINKED ISSUE #214: Docker Compose publishes FoundationDB on all host interfaces
## Summary

`docker/compose.yml` maps FoundationDB as `4500:4500`. Docker Compose port mappings without an explicit interface bind to `0.0.0.0` on the host by default. The same commit also commits `docker/vol/fdb/fdb.cluster` containing the connection string used by the Mosaic containers.

This is the reviewed outcome for Codex security finding [`c29efa2a`](https://chatgpt.com/codex/cloud/security/findings/c29efa2a5b148191add2c75577a8395b).

## Impact

- FoundationDB is Mosaic's durable state backend. Persisted state includes garbling/setup seeds and, after successful extraction, fault-secret material.
- A remote attacker with reachability to the host's port 4500 can connect with the public connection string (also committed in the repo) and read sensitive state, corrupt protocol state, or cause persistent denial of service.
- Severity: high (remote read/write of secret protocol state in any deployment that uses this Compose file as-is).

## Proposed fix

The Compose file is intended for container-internal use, so:

- prefer not publishing port 4500 to the host at all (containers can reach FDB over the Compose network)
- if a host port is needed for local debugging, bind it explicitly to loopback: `127.0.0.1:4500:4500`
- document that `docker/vol/fdb/fdb.cluster` is a development cluster file and must not be reused in production deployments

Also worth doing:

- add a comment in `compose.yml` explaining the loopback bind / no-bind choice so it doesn't get reverted
- make any documented "production" Compose example bind FDB only to the internal network and require TLS / network-policy controls

## Acceptance criteria

- `docker/compose.yml` does not expose port 4500 on a non-loopback host interface by default.

Zk2u:
## Summary

Closes #214.

The Compose-managed FoundationDB instance previously published `4500:4500`, which on Docker maps to `0.0.0.0:4500` on the host by default. Because the cluster file with the FDB connection string is also committed in the repo (`docker/vol/fdb/fdb.cluster`), anyone with reachability to the host's port 4500 could connect to the database using the public connection string and read, corrupt, or DoS protocol state — including persisted seed and fault-secret material.

Bind explicitly to `127.0.0.1:4500:4500`. Container-to-container traffic continues to use the Compose network at `foundationdb:4500` and is unaffected.

## Issues
Closes #214.

### DIFF
--- a/docker/compose.yml
+++ b/docker/compose.yml
@@ -18,8 +18,12 @@ services:
       - ./vol/fdb/fdb.cluster:/var/fdb/fdb.cluster:rw
       - ./vol/fdb/data:/var/fdb/data
     ports:
+      # Bind FoundationDB strictly to loopback so the database (which holds
+      # protocol-secret material) is reachable only from the local host.
+      # Inter-container traffic uses the Compose network at
+      # `foundationdb:4500` and is unaffected.
       # Keep this port in sync with ./vol/fdb/fdb.cluster
-      - "4500:4500"
+      - "127.0.0.1:4500:4500"
     healthcheck:
       test:
         [


sapinb (approved):
ACK a3750f15c65d94685e45ce2ccb73f9956c7999c5

Issue #212, opened 2026-05-03
Production Docker image ships reduced-circuits binary, selectable via env var


Zk2u:
## Summary

The production Docker image builds and includes both the normal `mosaic` binary and a `reduced-circuits` binary, and `docker/entrypoint.sh` selects the reduced binary at runtime based only on the `MOSAIC_REDUCED_CIRCUITS` environment variable.

The `reduced` feature lowers the cut-and-choose parameters from `N_CIRCUITS = 181` / `N_OPEN_CIRCUITS = 174` (documented as meeting the 40-bit target) to `N_CIRCUITS = 5` / `N_OPEN_CIRCUITS = 3` (~`log2(C(5,3)) = 3.3` bits of cut-and-choose). `crates/common/src/constants.rs` explicitly states these values are not secure and should only be used for testing.

This is the reviewed outcome for Codex security finding [`d8bac540`](https://chatgpt.com/codex/cloud/security/findings/d8bac5408c748191b2d65b18c3beb4bb).

## Impact

- An accidental or injected env-var setting in production deploys a node with materially weakened cut-and-choose soundness.
- A configured malicious peer would face only ~3 bits of challenge entropy, enabling realistic cheating attacks against garbled-circuit verification and downstream signing.
- The default path runs the normal binary, so the issue is purely a deploy-time footgun, but the probability is non-trivial because the switch is a single env var with no failure mode.
- Severity: high (production protocol soundness if triggered).

## Proposed fix

The reduced-circuits binary should not be in the production image at all.

Reasonable options:

- ship a separate test image (e.g. `mosaic-test`) with the reduced-circuits binary, and keep the production image single-binary
- or, if both must coexist, fail closed: refuse to start with the reduced binary unless an explicit "I understand this is insecure" config marker is also set, and log loudly at startup

Also worth doing regardless:

- have the reduced-circuits binary log a prominent warning at every startup
- name it differently from the production binary so it is obvious in process listings

## Acceptance criteria

- The default production Docker image does not contain a binary that can be activated into the reduced-circuits configuration via a single environment variable.

AaronFeickert:
This conflicts with the [specific decision](https://github.com/alpenlabs/mosaic/pull/170#issuecomment-4237384992) by @Zk2u to rely on an environment variable to avoid the need for separate images.

AaronFeickert:
[...]

PR #226, opened 2026-05-08
docker/entrypoint: require explicit confirmation for reduced-circuits binary

Zk2u:
## Summary

Refs #212.

The Docker image ships both `mosaic` (production) and `mosaic-unsafe-test` (built with `--features=reduced-circuits`) in the runtime image, with the entrypoint switching on `MOSAIC_UNSAFE_TEST`. The reduced-circuits build uses cut-and-choose parameters that are explicitly insecure (~3 bits of challenge entropy versus the 40-bit production target), so a single accidental env-var setting can flip a production deploy into a materially-weakened protocol-soundness state.

## Change

The entrypoint now requires **two** environment variables to enable the unsafe binary:

```
MOSAIC_UNSAFE_TEST=1
MOSAIC_UNSAFE_TEST_I_UNDERSTAND_THIS_IS_NOT_PRODUCTION=1
```

Setting only the first refuses to start with a clear error pointing to the second flag and the reason. When both are set, the entrypoint prints a loud multi-line warning before exec'ing the unsafe binary so the choice is visible in startup logs.

## Notes

This addresses the immediate footgun. The cleaner follow-up — splitting the unsafe binary into a separate Docker image so it is never present in production images at all — is tracked under the same issue (#212) and not closed by this PR.

CI environments that currently use `MOSAIC_UNSAFE_TEST=1` will need to add the second flag.

## Issues
Refs #212.

## Validation
- Manual entrypoint review (no Rust changes).

### DIFF
--- a/docker/entrypoint.sh
+++ b/docker/entrypoint.sh
@@ -8,8 +8,19 @@
 # The config file's [circuit] section should reference:
 #   path = "/etc/mosaic/circuit.v5c"
 #
-# Set MOSAIC_UNSAFE_TEST=1 to run in test mode, which is faster.
-# **WARNING**: Test mode is UNSAFE, and MUST NOT be used in production.
+# Test-mode binary (built with --features=reduced-circuits): for CI/test
+# environments only. This binary uses cut-and-choose parameters that are
+# explicitly documented as insecure (see crates/common/src/constants.rs)
+# and provides ~3 bits of challenge entropy versus the 40-bit production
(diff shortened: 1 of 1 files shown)


sapinb (approved):
ACK 160c8c29f1e77dd48820f470da03a6b23a4f2040

Although we should just remove the test binary entirely. It was originally only added to test deployments quickly, which is not necessary anymore.

Issue #211, opened 2026-05-03
ECR publish workflow allows pushing images built from arbitrary refs to production tags


Zk2u:
## Summary

The Docker → ECR publish workflow (`.github/workflows/docker-publish-ecr.yml`) accepts arbitrary `workflow_dispatch` inputs for both the source ref and `image_tag`. The selected ref is checked out, its Dockerfile is used as the build definition, and the resulting image is pushed to the production private ECR repository, then re-tagged and pushed to public ECR.

This is the reviewed outcome for Codex security finding [`5c9eaa8f`](https://chatgpt.com/codex/cloud/security/findings/5c9eaa8f079c81919681bd93ae14e65e).

## Impact

- A workflow dispatcher can build an image from an unreviewed branch/commit and push it to a meaningful tag (e.g. `latest` or a release tag), overwriting the production-trusted image.
- Concretely a supply-chain integrity issue when workflow-dispatch permission is broader than production release permission, which is common on GitHub repos that don't gate on `environments`.
- Severity: high (production image integrity).

## Proposed fix

Constrain who and what can push to production tags. Reasonable options:

- gate the push job on a GitHub `environment` with required reviewers
- restrict allowed `image_tag` values (e.g. only commit SHAs of the checked-out ref, no `latest`/release-tag overwrites from dispatch)
- restrict allowed source refs (e.g. only protected branches and signed tags)
- separate "build" from "publish-to-production" so the publish step requires explicit review even if build is broadly available

## Acceptance criteria

- It is not possible for a user with repo write access alone to overwrite a production-meaningful image tag in ECR by dispatching the workflow against an arbitrary ref.

PR #227, opened 2026-05-08
ci(docker-publish-ecr): refuse reserved image tags from workflow_dispatch

Zk2u:
## Summary

Refs #211.

The Docker → ECR publish workflow accepts arbitrary `ref` and `image_tag` inputs. A workflow dispatcher can therefore build an image from an unreviewed branch or commit and push it to a production-meaningful tag, overwriting the trusted image at that tag.

This PR rejects `image_tag` values that could overwrite production-meaningful tags when supplied via `workflow_dispatch`:

- Exact reserved names: `latest`, `main`, `master`, `stable`, `prod`, `production`, `release`
- Release-shaped tags: `v?N.M[.K][...]` (matches `v1.2.3`, `1.0`, etc.)
- Release-prefixed tags: `release-*`, `release_*`, `release.*`

The default tagging path (resolved short SHA) is unchanged. Release tags must come from a separate, reviewed release workflow rather than from a manual dispatch against an arbitrary ref.

## Notes

This is the in-YAML mitigation. Stronger gating — a GitHub `environment` with required reviewers on the AWS environment — is an org-level configuration change tracked under #211 and not addressed by this PR. Doing both gives defense-in-depth.

## Issues
Refs #211.

## Validation
- Workflow file YAML is unchanged outside the validate step; the bash logic uses standard `bash` constructs already used elsewhere in the file.

### DIFF
--- a/.github/workflows/docker-publish-ecr.yml
+++ b/.github/workflows/docker-publish-ecr.yml
@@ -71,6 +71,27 @@ jobs:
               echo "::error::image_tag must match [A-Za-z0-9._-]{1,128}"
               exit 1
             fi
+
+            # Refuse user-supplied image_tag values that could overwrite
+            # production-meaningful tags from an arbitrary ref. The default
+            # path (no override) tags by short SHA, which cannot collide.
+            #
+            # Reserved exact tags:
+            case "${INPUT_IMAGE_TAG}" in
+              latest|main|master|stable|prod|production|release)
+                echo "::error::image_tag '${INPUT_IMAGE_TAG}' is reserved; not allowed via workflow_dispatch"
+                exit 1
+                ;;
+            esac
+            # Reserved release-shaped tags (e.g. v1.2.3, 1.2.3, release-*).
+            if [[ "${INPUT_IMAGE_TAG}" =~ ^v?[0-9]+\.[0-9]+(\.[0-9]+)?([._-].*)?$ ]]; then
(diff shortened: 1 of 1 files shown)


sapinb:
@madan-oss Does this impact deployment ?

sapinb:
@Zk2u Instead of blacklisting specific edgecases, can we only allow the commit sha ref path ? 
Something like
```
# Only allow hex strings (commit SHAs) as image tags from
# workflow_dispatch. This prevents overwriting production-
# meaningful tags (latest, semver, release-*) from arbitrary refs.
if [[ ! "${INPUT_IMAGE_TAG}" =~ ^[0-9a-f]{8,40}$ ]]; then
  echo "::error::image_tag must be a hex commit SHA (8-40 chars); got '${INPUT_IMAGE_TAG}'"
  exit 1
fi
```

Zk2u:
We can, that said that means updating images for operators is one step more complex. Madan's decision

madan-oss:
> @madan-oss Does this impact deployment ?

LGTM. Fail-fast guard is correct and fills the gap ECR IMMUTABLE can't cover (new tags that don't exist yet). One optional suggestion: normalize `INPUT_IMAGE_TAG` to lowercase before the `case` block so `LATEST`/`PROD`/`Main` are also caught. Fine to merge as-is or pick that up as a follow-up.

AaronFeickert:
@Zk2u: blocking on the normalization suggestion from @madan-oss, in addition to the signed commit requirement for this repository.

AaronFeickert (approved):

PR #229, opened 2026-05-08
cac/protocol,vs3: validate challenge indices and Index canonical-deser

### LINKED ISSUE #189: Challenge indices should be validated before garbler STF uses them
## Summary

Challenge indices are deserialized without real bounds validation, and garbler-side code later indexes arrays directly and can panic on out-of-range or duplicate indices.

This is the reviewed outcome for audit finding `#48180`.

## Impact

- Authenticated peer can trigger STF panic / DoS with malformed challenge indices.
- Under our operating model, this is low severity, but it is still a real panic path.

## Proposed fix

Validate challenge indices at the protocol boundary before storing or acting on them.

Checks should include:
- index range bounds
- duplicate detection
- reserved-index rules
- exact expected challenge count if applicable

Also remove panic-based assumptions in downstream helper paths.

## Acceptance criteria

- Malformed challenge indices result in a normal protocol rejection/abort path, not a panic.

Zk2u:
## Summary

Closes #189.

Two related defenses against malformed peer-supplied challenge indices:

### 1. `Valid::check` for `Index` was a no-op

`Index::CanonicalDeserialize` casts a wire `u64` to `usize` without bounds checking, and `Valid::check` returned `Ok(())` unconditionally. A peer can produce indices outside `[MIN, MAX]` or the reserved value (0), and any canonical deserializer configured with `Validate::Yes` accepted them.

The `Valid::check` impl now rejects out-of-range and reserved values, so deserializers fail closed instead of handing downstream code an out-of-range `Index` that later panics in array indexing or fixed-length conversion.

### 2. `is_valid_challenge` did not detect duplicates

`is_valid_challenge` checked only that no entry equalled the reserved index. Without a duplicate check, `get_eval_indices` (which builds a `[Index; N_EVAL_CIRCUITS]` via `try_into`) panics when the challenge contains duplicates, because the unchallenged set has the wrong cardinality. That is a DoS path reachable from any authenticated peer.

`is_valid_challenge` now enforces:

- no entry equals `Index::reserved()`
- every entry is in `[Index::MIN, Index::MAX]`
- no duplicates

## Tests

Three new tests in `garbler::stf::challenge_validation_tests`:

- `rejects_reserved_index`
- `rejects_duplicate_indices`
- `accepts_well_formed_challenge`

## Issues
Closes #189.

## Validation
- `cargo check --workspace`
- `cargo test -p mosaic-cac-protocol`

### DIFF
--- a/crates/cac/protocol/src/garbler/stf.rs
+++ b/crates/cac/protocol/src/garbler/stf.rs
@@ -1110,15 +1110,36 @@ fn generate_garbling_table_seeds(base_seed: Seed) -> AllGarblingSeeds {
     HeapArray::from_vec(garbling_seeds)
 }
 
-/// challenge indices must be in range, must not include 0, etc
+/// Validate a peer-supplied [`ChallengeMsg`].
+///
+/// `ChallengeIndices` is a fixed-length `HeapArray<Index, N_OPEN_CIRCUITS>`,
+/// so cardinality is implicit. We additionally enforce:
+///
+/// - no entry equals the reserved index (0),
+/// - every entry is in `[Index::MIN, Index::MAX]` (defense in depth — the `Valid` impl on `Index`
+///   also rejects out-of-range values during canonical deserialization, but we don't rely on the
+///   caller having opted in to `Validate::Yes`),
+/// - no duplicates.
+///
+/// Without the duplicate check, downstream code paths (`get_eval_indices`'s
+/// fixed-length conversion in particular) panic on malformed input from an
+/// authenticated peer, which is a DoS vector.
 fn is_valid_challenge(challenge: &ChallengeMsg) -> bool {
-    !challenge
-        .challenge_indices
-        .iter()
-        .any(|x| *x == Index::reserved())
-    // does not include reserved index
-    // ChallengeMsg in itself includes `Index` struct which can only be initialized within valid
-    // bounds so the range check is done during deserialization itself
+    let reserved = Index::reserved();
+    let mut seen = std::collections::HashSet::with_capacity(challenge.challenge_indices.len());
+    for idx in challenge.challenge_indices.iter() {
+        if *idx == reserved {
+            return false;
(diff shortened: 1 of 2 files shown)


Zk2u:
@codex rereview please

Zk2u:
## Merge dependency

This PR is a downstream of **#222** (unbound internal channels). #222 must land first.

The functional-test failures observed on this PR (in particular setup-phase timeouts in `fn_mosaic_setup_*` / `fn_mosaic_deposit_*` / `fn_mosaic_withdrawal_concurrent`) are the bounded-queue cycle deadlock that #222 fixes (filed as **#221**). The PR's own changes are independent of that deadlock — once #222 is in main and this branch is rebased, the FT failures clear.

Suggested merge order: #222 → rebase this branch on main → re-run CI → merge.

Zk2u:
@codex rereview please

Zk2u:
@codex rereview please

sapinb (approved):
ACK 59c20486a43229e1d1faed17c337296509f81b65

crates/cac/protocol/src/garbler/stf.rs
```diff
@@ -1173,3 +1194,50 @@ fn get_eval_seeds(
...
+
+    use super::*;
+
+    fn make_challenge(indices: &[usize]) -> ChallengeMsg {
+        let challenge_indices: ChallengeIndices = HeapArray::new(|i| {
+            let raw = indices[i];
+            // Allow constructing out-of-range entries for negative tests by
+            // bypassing `Index::new`. The Index API does not normally permit
+            // 0 / out-of-range values, but a malicious peer can produce them
+            // by hand-crafting wire bytes; the validator must catch that.
```
AaronFeickert, line 1211: @Zk2u this isn't what happens here for non-reserved indexes. In fact, we don't test the non-reserved out-of-range case at all.

PR #230, opened 2026-05-08
job/executors: chunk evaluator share verification to bound peak memory

### LINKED ISSUE #216: Batch share verification allocates O(N_INPUT_WIRES * 256 * N_OPEN_CIRCUITS) points before rejecting invalid shares
## Summary

The evaluator's `VerifyOpenedInputShares` path materializes the full batch of opened shares before any verification step:

- transposes all opened shares into `share_bufs`, duplicating ~`N_INPUT_WIRES * WIDE_LABEL_VALUE_COUNT * N_OPEN_CIRCUITS` `Share` objects
- `batch_verify_shares` then copies every coefficient point and matching scalar into `points` / `scalars` vectors
- `Point::normalize_batch` produces another full affine vector while the originals are still live

With production constants (`N_INPUT_WIRES = 164`, `WIDE_LABEL_VALUE_COUNT = 256`, `N_OPEN_CIRCUITS = 174`) this is on the order of `164 × 256 × 175 ≈ 7.3M` projective points plus matching scalars in a single allocation pass, in addition to the already-loaded shares and commitments.

This is the reviewed outcome for Codex security finding [`247fa114`](https://chatgpt.com/codex/cloud/security/findings/247fa114252081919168055f89a95e13).

## Impact

- Invalid shares are only rejected after these allocations.
- A configured malicious peer can send well-sized challenge-response chunks that pass earlier framing/length checks but fail batch verification, forcing the full `7M+` allocation each round.
- Likely outcome on production hardware is severe memory pressure or OOM under repeated attempts.
- The previous loop verified shares one at a time and did not introduce these full-batch materializations.
- Severity: high (resource exhaustion via authenticated peer; reachable on any node that performs evaluator-side share verification).

## Proposed fix

Avoid materializing the full batch as a precondition for verification.

Reasonable options:

- chunk `batch_verify_shares` into bounded windows (e.g. one wire group at a time) so peak memory is `O(window)` instead of `O(total)`
- restore per-share verification at least as a fallback path until a memory-bounded batch verifier is available
- if batch verification has measurable correctness or performance value, gate the "full batch" path on a sanity admission check that bounds total points/scalars allocated

Also worth doing:

- add an explicit upper bound on the per-job allocation budget for evaluator share verification, so this kind of regression fails closed in the future
- consider reusing buffers across calls so we don't double-allocate via `normalize_batch`

## Acceptance criteria

- A configured peer cannot drive evaluator memory use proportional to the full opened-share batch on every attempt.
- Peak memory of share verification is bounded by a configurable window rather than the full opened-share set.

Zk2u:
## Summary

Closes #216.

`handle_verify_opened_input_shares` previously materialized every opened share and every `(commitment, share)` pair before any verification ran. With production constants (`N_INPUT_WIRES × WIDE_LABEL_VALUE_COUNT × N_OPEN_CIRCUITS = 164 × 256 × 174 ≈ 7.3M` shares, plus matching point/scalar vectors of similar size), this is a multi-million-point allocation pass that runs on every challenge response, including ones that will eventually fail — a memory-pressure DoS reachable from an authenticated peer who can send well-sized but invalid `ChallengeResponseMsgChunk` payloads.

## Change

Process the verification in fixed-size chunks (256 pairs at a time). Each chunk runs an independent RLC `batch_verify_shares` call. Per-chunk soundness error is `chunk_size / |F|`, still negligible for a 256-bit field, so this preserves correctness while keeping peak memory bounded by chunk size rather than full input.

`VERIFY_CHUNK_PAIRS = 256` chosen as a default that comfortably fits in MBs at production sizes; tunable.

## Issues
Closes #216.

## Validation
- `cargo check --workspace`
- `cargo test -p mosaic-job-executors --lib` (10 lib tests pass)
- `cargo test -p mosaic-vs3 batch` (7 batch-verify tests pass)

### DIFF
--- a/crates/job/executors/src/evaluator.rs
+++ b/crates/job/executors/src/evaluator.rs
@@ -171,33 +171,58 @@ pub(crate) async fn handle_verify_opened_input_shares<SP: StorageProvider, TS: T
         return HandlerOutcome::Retry;
     };
 
-    // Batch-verify all opened shares against their polynomial commitments via RLC.
-    // Collects (commitment, shares) pairs and verifies in a single MSM.
-    #[allow(clippy::needless_range_loop)]
+    // Batch-verify opened shares against their polynomial commitments in
+    // bounded chunks. The previous implementation materialized all
+    // `N_INPUT_WIRES * WIDE_LABEL_VALUE_COUNT * N_OPEN_CIRCUITS` shares plus
+    // a matching point/scalar vector before any verification ran, which
+    // under production constants is on the order of several million points
+    // and is a memory-pressure DoS vector for an authenticated peer (the
+    // verifier path is reachable with peer-supplied shares that fail late).
+    //
+    // Chunking preserves correctness: `batch_verify_shares` already uses an
+    // RLC-based check whose soundness error is `N / |F|`; running it on
+    // independent windows just yields a per-window check at the same
+    // soundness, with peak memory bounded by the chunk size.
     let failure_reason = {
-        let mut share_bufs: Vec<Vec<Share>> =
-            Vec::with_capacity(N_INPUT_WIRES * WIDE_LABEL_VALUE_COUNT);
+        // Tunable. Each chunk holds roughly
+        //   VERIFY_CHUNK_PAIRS * (N_OPEN_CIRCUITS shares + N_COEFFICIENTS points + scalars)
+        // so a chunk of 256 pairs is on the order of MBs at production sizes.
+        const VERIFY_CHUNK_PAIRS: usize = 256;
+        let total_pairs = N_INPUT_WIRES * WIDE_LABEL_VALUE_COUNT;
+        let mut rng = rand::rngs::OsRng;
+        let mut result: Option<String> = None;
 
-        for wire in 0..N_INPUT_WIRES {
-            for val in 0..WIDE_LABEL_VALUE_COUNT {
-                let wire_val_shares: Vec<Share> = (0..N_OPEN_CIRCUITS)
+        let mut chunk_shares: Vec<Vec<Share>> = Vec::with_capacity(VERIFY_CHUNK_PAIRS);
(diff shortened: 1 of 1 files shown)


nakkstar123:
This was added by me as a memory-runtime tradeoff. I don't think batching scales that well for small chunks like 256 shares. Can you check how this change affects runtime (of specifically opened input share verification)?

nakkstar123:
Further, I do not think this is vulnerable to a DoS attack. There isn't scope for memory exhaustion, it's a fixed, bounded (although large) memory required for batch verification. If any share doesn't verify, the check fails and the protocol aborts.

AaronFeickert:
> This was added by me as a memory-runtime tradeoff. I don't think batching scales that well for small chunks like 256 shares. Can you check how this change affects runtime (of specifically opened input share verification)?

Any update on this?

crates/job/executors/src/evaluator.rs
```diff
@@ -163,33 +163,58 @@ pub(crate) async fn handle_verify_opened_input_shares<SP: StorageProvider, TS: T
...
+    //
+    // Chunking preserves correctness: `batch_verify_shares` already uses an
+    // RLC-based check whose soundness error is `N / |F|`; running it on
+    // independent windows just yields a per-window check at the same
+    // soundness, with peak memory bounded by the chunk size.
     let failure_reason = {
-        let mut share_bufs: Vec<Vec<Share>> =
-            Vec::with_capacity(N_INPUT_WIRES * WIDE_LABEL_VALUE_COUNT);
+        // Tunable. Each chunk holds roughly
+        //   VERIFY_CHUNK_PAIRS * (N_OPEN_CIRCUITS shares + N_COEFFICIENTS points + scalars)
```
nakkstar123, line 188: This looks good to me. Previously, I mistakenly thought that exactly 256 shares were in one chunk, but it's one of the variables in the total calculation that's being scaled down to 256.

nakkstar123 (approved):

PR #231, opened 2026-05-08
storage/fdb: stream read-only ranges instead of materializing them

### LINKED ISSUE #185: Read-only FDB deposit scans eagerly materialize full ranges
## Summary

The read-only FoundationDB range path eagerly collects the entire range before yielding the first item. RPC deposit listing and restart-time restore both iterate all deposits through that read-only path.

This is the reviewed outcome for audit finding `#48102`.

## Impact

- Large deposit sets can cause allocation spikes during RPC listing and node restart/restore.
- This is specific to the read-only FDB path.
- Under our operating model, this is a medium-severity availability issue.

## Proposed fix

Make read-only FDB range traversal actually incremental.

Implementation work:
- replace the eager `collect_range()`-based implementation with a paged/streamed cursor that yields items progressively
- audit callers that still do eager `try_collect()` and either paginate externally or keep full-collection semantics only where the result set is intentionally bounded

## Acceptance criteria

- Read-only FDB deposit iteration does not allocate the full range up front before the first item is yielded.

Zk2u:
## Summary

Closes #185.

`FdbReadOnlyStore::range` previously called `collect_range`, which opened a single `transact_boxed` and accumulated every key-value pair across all paginated `get_range` calls into a `Vec<KvPair>` before returning. Callers like the deposit-listing RPC and restart-time restore both iterate this stream, so a node with a large deposit population paid an O(N) allocation up front before the first item became visible.

## Change

Replace the eager collection with a true page-by-page stream:

- `range` opens a fresh `Transaction` lazily inside the stream future.
- Items are yielded from the in-memory page; once that page is drained, the next page is fetched on demand against the same transaction.
- The transaction is captured by `Arc<Transaction>` so it lives as long as the stream and is dropped when the stream is dropped.

Removes the now-unused `collect_range`, `stream_from_pairs`, `RangeRequest`, and standalone `make_cursor` helpers.

## Trade-off

`create_trx()` is used directly rather than `transact_boxed`, so the read does not auto-retry on transaction conflicts. For read-only iteration this is acceptable: read-only transactions don't conflict in FDB's MVCC model. If a transaction times out (FDB's 5s soft limit) the stream surfaces the error normally.

## Issues
Closes #185.

## Validation
- `cargo check --workspace`
- `cargo test -p mosaic-storage-fdb` (3 tests pass)

### DIFF
--- a/crates/storage/fdb/src/lib.rs
+++ b/crates/storage/fdb/src/lib.rs
@@ -7,7 +7,9 @@
 use std::{
     collections::{HashMap, VecDeque},
     fmt,
+    future::Future,
     ops::Bound,
+    pin::Pin,
     sync::{Arc, Mutex},
 };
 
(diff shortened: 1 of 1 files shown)


Zk2u:
@codex rereview please

Zk2u:
@codex rereview please

Zk2u:
## Merge dependency

This PR is a downstream of **#222** (unbound internal channels). #222 must land first.

The functional-test failures observed on this PR (in particular setup-phase timeouts in `fn_mosaic_setup_*` / `fn_mosaic_deposit_*` / `fn_mosaic_withdrawal_concurrent`) are the bounded-queue cycle deadlock that #222 fixes (filed as **#221**). The PR's own changes are independent of that deadlock — once #222 is in main and this branch is rebased, the FT failures clear.

Suggested merge order: #222 → rebase this branch on main → re-run CI → merge.

sapinb (approved):
ACK d31c38a4c3e8b82fc7ac7d90c8185a97b45ed17c

PR #173, opened 2026-04-17
add mosaic ecr publish workflow

madan-oss:
## Description

Adds a manual `workflow_dispatch` workflow to build and publish the Mosaic Docker image to both private and public ECR, with Trivy vulnerability scanning rendered as a Markdown table in the workflow summary page.

**What the workflow does:**

- Accepts optional `ref` and `image_tag` inputs with strict allowlist validation
- Checks out the selected ref, resolves final SHA and image tag
- Builds `docker/Dockerfile` for `linux/amd64` with Docker Buildx + GHA layer cache for faster Rust rebuilds
- Pushes to private ECR (`888577024788.dkr.ecr.us-east-1.amazonaws.com/tn1/mosaic`) via OIDC-assumed role
- Re-tags and pushes to public ECR (`public.ecr.aws/z5c7y9u9/mosaic`) via a separate public ECR OIDC role using `docker tag + push` (not `imagetools create` — avoids ECR credential helper re-auth issues at push time)
- Resolves the pushed image digest from private ECR
- Runs Trivy twice: JSON scan → rendered as a Markdown severity table in the workflow summary; SARIF scan → uploaded to the GitHub Security tab. Both are advisory (`exit-code: 0`) and never block the publish
- Appends image references, digest, and Trivy table to `$GITHUB_STEP_SUMMARY`

**Files added:**

| File | Purpose |
| --- | --- |
| `.github/workflows/docker-publish-ecr.yml` | Main publish workflow |
| `.github/scripts/render_trivy_summary.py` | Converts `trivy-results/mosaic.json` → Markdown table for the workflow summary page |

### Type of Change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature/Enhancement (non-breaking change which adds functionality or enhances an existing one)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency update
- [ ] Security fix

## Notes to Reviewers

- AWS auth is OIDC throughout — no long-lived credentials anywhere in the workflow.
- Two AWS roles are required (both must exist before this workflow can run):
  - Private ECR: `AWS_ROLE_TO_ASSUME` repo variable → `arn:aws:iam::888577024788:role/github-actions-mosaic-ecr-publish`
[...]

PR #253, opened 2026-05-15
ci: enforce Trivy as hard gate before ECR push

madan-oss:
## Description

Implements STR-2637. Adds Trivy vulnerability scanning as a hard gate in the ECR publish pipeline — no image reaches ECR unless it passes a CRITICAL/HIGH scan.

Builds on the two-checkout isolation introduced in #173 and addresses the one security gap that remained in that PR.

### Type of Change

- [x] Security fix

## What changed

**Pipeline order** — previously the image was pushed first, then scanned. Scan now runs against the locally built image before any push. A failed gate means ECR is never touched.

| Step | Before | After |
|---|---|---|
| Build | `--push` (immediate ECR push) | `--load` (local daemon only) |
| Hard gate | advisory, post-push | blocking, pre-push, `exit-code: 1` |
| SARIF | advisory post-push | advisory pre-push, `if: always()` so results appear even on gate failure |
| SBOM (CycloneDX) | not present | advisory, attached as artifact |
| Push to ECR | immediate on build | only after gate passes |

**Severity** — gate blocks on `CRITICAL,HIGH` with `ignore-unfixed: true`. SARIF captures `CRITICAL,HIGH,MEDIUM,LOW` for full visibility in the Security tab without blocking.

**`.trivyignore`** — anchored to the root checkout (always `main`). Only reviewed, merged code can suppress CVEs. An untrusted `inputs.ref` cannot manipulate the ignore list — this closes the bypass vector noted in #173's review.

**Branch trigger** — added `push: branches: ['infra/**']` with path filter so the pipeline is exercisable on feature branches without a manual dispatch.

**Supporting files**
- `.trivyignore` — empty with format docs; add justified exceptions here via PR
- `.github/scripts/render_trivy_summary.py` — renders the JSON report as a Markdown table in the workflow summary
- `docs/trivy-scanning.md` — how to read results and add CVE exceptions

## Notes to Reviewers

The `trivyignores` path was intentionally kept at `.trivyignore` (root/main checkout) rather than `build-src/.trivyignore`. This mirrors the same trust boundary established in #173 for workflow scripts — the ignore list requires a reviewed PR to change, preventing an untrusted `inputs.ref` from suppressing real CVEs.

Three CI runs on this branch validated the full flow:
1. Gate fired correctly on zero-unfixed-CVE image → passed, image pushed ✅
2. Gate fired correctly when MEDIUM/LOW gate was tested → blocked, no push ✅
3. Final config (CRITICAL/HIGH gate, all-severity SARIF) → passed, image pushed ✅

## Checklist

## Related Issues

closes STR-2637

### DIFF
--- a/.github/scripts/render_trivy_summary.py
+++ b/.github/scripts/render_trivy_summary.py
@@ -1,122 +1,58 @@
 #!/usr/bin/env python3
-"""render_trivy_summary.py — Convert a Trivy JSON report to a Markdown summary table.
-
-Called by the "Render Trivy markdown summary" step in docker-publish-ecr.yml after
-the JSON-format Trivy scan completes. Reads trivy-results/mosaic.json and writes
-trivy-results/mosaic.md, which is then appended to $GITHUB_STEP_SUMMARY so the
-table appears on the workflow summary page.
-
-A separate SARIF-format scan uploads structured results to the GitHub Security tab
-for tracking; this script handles only the human-readable workflow summary.
-
-Why JSON → Python instead of the built-in Trivy table format?
-  The Trivy `table` format emits plain text which renders as a code block in
-  $GITHUB_STEP_SUMMARY. JSON → Python lets us produce a proper Markdown table
-  (rendered natively by GitHub) that matches the strata-bridge workflow layout.
-
-Expected env vars (injected by the workflow step's env: block):
-    IMAGE_REF  — full ECR image reference shown in the summary header
-    IMAGE_TAG  — short SHA tag (unused in output, available for future use)
-
-Output:
-    trivy-results/mosaic.md
-"""
+"""Parse trivy-results/mosaic.json and write a Markdown summary to trivy-results/mosaic.md."""
 
 import json
 import os
-from collections import Counter
+import sys
 from pathlib import Path
 
+IMAGE_REF = os.environ.get("IMAGE_REF", "unknown")
+IMAGE_TAG = os.environ.get("IMAGE_TAG", "unknown")
 
-# Canonical ordering used for the severity count table columns and finding sort.
-SEVERITY_ORDER = ["CRITICAL", "HIGH", "MEDIUM", "LOW", "UNKNOWN"]
-
-# Maximum number of individual findings to list in the summary.
-# Keeps the summary page scannable — full results are in the uploaded artifact.
-MAX_FINDINGS = 10
+INPUT = Path("trivy-results/mosaic.json")
+OUTPUT = Path("trivy-results/mosaic.md")
 
 
-def iter_items(value: object) -> list[dict]:
-    """Normalise a Trivy result field to a flat list of dicts.
+def main():
+    if not INPUT.exists():
+        print(f"::warning::Trivy JSON not found at {INPUT}, skipping summary", file=sys.stderr)
+        return
 
-    Trivy can emit a list, a dict, or null for the same field depending on the
-    scanner mode and image content. This handles all three cases without fragile
-    shell conditionals.
-    """
-    if isinstance(value, list):
-        return [item for item in value if isinstance(item, dict)]
-    if isinstance(value, dict):
-        return [item for item in value.values() if isinstance(item, dict)]
-    return []
+    data = json.loads(INPUT.read_text())
 
+    vulns = []
+    for result in data.get("Results", []):
+        for v in result.get("Vulnerabilities") or []:
+            vulns.append({
+                "severity": v.get("Severity", ""),
+                "id": v.get("VulnerabilityID", ""),
+                "pkg": v.get("PkgName", ""),
+                "installed": v.get("InstalledVersion", ""),
+                "fixed": v.get("FixedVersion", "") or "—",
+                "title": v.get("Title", ""),
+            })
 
-def main() -> None:
-    image_ref = os.environ.get("IMAGE_REF", "mosaic")
-    json_path = Path("trivy-results/mosaic.json")
+    lines = [f"### Trivy vulnerability scan — `{IMAGE_TAG}`", ""]
 
-    lines: list[str] = [
-        "### Trivy Summary: Mosaic",
-        "",
-        f"`{image_ref}`",
-        "",
-        "- Scanners: `vuln`",
-        "- Severity filter: `HIGH,CRITICAL` (unfixed vulnerabilities ignored)",
-        "",
-    ]
-
-    if not json_path.exists():
-        # JSON scan step may have been skipped or failed; emit a placeholder rather
-        # than crashing the summary step.
-        lines += ["Trivy did not produce a JSON report for this image.", ""]
+    if not vulns:
+        lines.append(f":white_check_mark: No HIGH or CRITICAL vulnerabilities found in `{IMAGE_REF}`.")
     else:
-        results = json.loads(json_path.read_text(encoding="utf-8"))
-        counts: Counter = Counter()
-        findings: list[dict] = []
-
-        for target in results.get("Results", []):
-            if not isinstance(target, dict):
-                continue
-            target_name = target.get("Target", "unknown")
-
-            for vuln in iter_items(target.get("Vulnerabilities")):
-                sev = vuln.get("Severity", "UNKNOWN")
-                counts[sev] += 1
-                findings.append({
-                    "severity": sev,
-                    "target": target_name,
-                    "kind": "vuln",
-                    "id": vuln.get("VulnerabilityID", "unknown"),
-                })
-
-        # Sort CRITICAL → HIGH → MEDIUM → LOW → UNKNOWN, then alphabetically by ID
-        # for stable ordering across runs.
-        findings.sort(key=lambda f: (
-            SEVERITY_ORDER.index(f["severity"]) if f["severity"] in SEVERITY_ORDER else len(SEVERITY_ORDER),
-            f["id"],
-        ))
-
-        # Severity count summary table
         lines += [
-            "| CRITICAL | HIGH | MEDIUM | LOW | UNKNOWN |",
-            "| --- | --- | --- | --- | --- |",
-            "| " + " | ".join(str(counts.get(s, 0)) for s in SEVERITY_ORDER) + " |",
+            f":x: **{len(vulns)} HIGH/CRITICAL vulnerability(ies)** found in `{IMAGE_REF}`.",
             "",
+            "| Severity | CVE | Package | Installed | Fixed | Summary |",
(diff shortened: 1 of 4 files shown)


Zk2u (commented):
Review pass. Solid PR overall — the hard-gate-before-push reordering is exactly the right shape, the two-checkout trust boundary for `.trivyignore` is correctly preserved, and the SARIF/SBOM advisory pattern with `if: always()` is sound. A few things worth surfacing before merge:

## Blockers / questions

**1. Trigger doesn't match the PR description.** The body says:

> added `push: branches: ['infra/**']` with path filter so the pipeline is exercisable on feature branches without a manual dispatch

The actual YAML has `push: branches: [main]` with no path filter (see commit `be8878f ci: trigger on push to main for automatic CD`, which superseded the earlier `09ad419 ci: remove push trigger`). This is a meaningful policy change buried under a scope item described as "branch-trigger for testing":

- Pre-PR: every ECR publish was `workflow_dispatch` (explicit, manual, audit-friendly).
- Post-PR: every push to main auto-publishes.

Is the automatic-CD-from-main behaviour intentional for this PR, or was the description correct and the trigger should be `infra/**` (or excluded entirely until env-reviewer config from #211 lands)? Worth either updating the description to reflect "auto-publish on push to main" or reverting the trigger to match the description.

Related: no `paths-ignore` filter, so doc-only changes and PRs that only touch unrelated crates will also rebuild + publish.

**2. Auto-publish from main relies on the `AWS` environment having required reviewers configured.** The job has `environment: AWS`, so if (and only if) the env has deployment-protection reviewers, the auto-publish blocks on approval. This is the same setting #211 is tracking. If those reviewers aren't configured today, this PR effectively enables fully-automatic publishing of every main push. Worth explicitly confirming the env config is in place before merging this.

## Minor

**3. `render_trivy_summary.py` is largely rewritten (`+58/-122`).** The new version is leaner but drops:
- The severity-count summary table (CRITICAL/HIGH/MEDIUM/LOW/UNKNOWN row).
- The MAX_FINDINGS=10 truncation safeguard — all findings now render verbatim, which could blow up the workflow summary if a CVE-heavy base image lands.
- Multi-target enumeration (the new version flattens vulns across all `Results` entries, which is fine for image scans but loses the per-target column).
- The header docstring explaining the JSON→Python rationale (still useful context).

Not blocking, but worth knowing this is a behavioural change to the summary UX in addition to the gate change. If the team is OK with the simpler layout, fine.

**4. SBOM-as-artifact, not OCI-attached.** The CycloneDX SBOM is uploaded as a workflow artifact (90-day default retention). For supply-chain provenance, attaching the SBOM to the image via `cosign attach sbom` (or buildx provenance attestation) is the more durable pattern — survives artifact expiration and travels with the image. Out of scope for this PR but worth a follow-up.

**5. `scanners: vuln` on the SBOM step.** Technically harmless — CycloneDX is engaged by `format`, and `scanners: vuln` embeds vulnerability metadata in the SBOM. If that embedding is intentional (good for supply-chain), worth a comment saying so. If not, can drop.

**6. Earlier suppression commits (`99480ef ci: suppress 3 MEDIUM base image CVEs`) are superseded by the CRITICAL/HIGH-only gate.** Confirmed `.trivyignore` is empty in the final diff — good. Those MEDIUM CVEs will still appear in the Security tab via the all-severity SARIF scan, which is the right outcome.

**7. Trivy SARIF scan re-runs the full vulnerability analysis** (`severity: CRITICAL,HIGH,MEDIUM,LOW` vs hard-gate's `CRITICAL,HIGH`). Two scans of the same image is duplicate work, but Trivy's DB cache makes it cheap. Not worth optimising unless CI time becomes a concern.

## Verified clean

- `.trivyignore` correctly resolves via the root (main) checkout, not `build-src/.trivyignore` — closes the bypass vector noted in #173. Verified by re-reading the two-checkout layout in the workflow.
- All Trivy action invocations pinned by SHA + version tag.
- Push step has no `if: always()` — it skips on gate failure as expected.
- Concurrency group correctly serialises by ref; push and dispatch converge on the same key when both target main.
- `continue-on-error: true` on advisory steps doesn't mask the hard gate (which lacks it).
- `if: always()` ordering means SARIF + SBOM + summary artifact upload all run on gate failure, so even blocked builds produce diagnostics.

## Suggested action

Decide intentionally on point 1 (trigger scope) and confirm point 2 (env-reviewer config in place), then this is good to merge. The Trivy gating itself is well-structured.

madan-oss:
> Review pass. Solid PR overall — the hard-gate-before-push reordering is exactly the right shape, the two-checkout trust boundary for `.trivyignore` is correctly preserved, and the SARIF/SBOM advisory pattern with `if: always()` is sound. A few things worth surfacing before merge:
> 
> ## Blockers / questions
> **1. Trigger doesn't match the PR description.** The body says:
> 
> > added `push: branches: ['infra/**']` with path filter so the pipeline is exercisable on feature branches without a manual dispatch
> 
> The actual YAML has `push: branches: [main]` with no path filter (see commit `be8878f ci: trigger on push to main for automatic CD`, which superseded the earlier `09ad419 ci: remove push trigger`). This is a meaningful policy change buried under a scope item described as "branch-trigger for testing":
> 
> * Pre-PR: every ECR publish was `workflow_dispatch` (explicit, manual, audit-friendly).
> * Post-PR: every push to main auto-publishes.
> 
> Is the automatic-CD-from-main behaviour intentional for this PR, or was the description correct and the trigger should be `infra/**` (or excluded entirely until env-reviewer config from #211 lands)? Worth either updating the description to reflect "auto-publish on push to main" or reverting the trigger to match the description.
> 
> Related: no `paths-ignore` filter, so doc-only changes and PRs that only touch unrelated crates will also rebuild + publish.
> 
> **2. Auto-publish from main relies on the `AWS` environment having required reviewers configured.** The job has `environment: AWS`, so if (and only if) the env has deployment-protection reviewers, the auto-publish blocks on approval. This is the same setting #211 is tracking. If those reviewers aren't configured today, this PR effectively enables fully-automatic publishing of every main push. Worth explicitly confirming the env config is in place before merging this.
> 
> ## Minor
> **3. `render_trivy_summary.py` is largely rewritten (`+58/-122`).** The new version is leaner but drops:
> 
> * The severity-count summary table (CRITICAL/HIGH/MEDIUM/LOW/UNKNOWN row).
> * The MAX_FINDINGS=10 truncation safeguard — all findings now render verbatim, which could blow up the workflow summary if a CVE-heavy base image lands.
> * Multi-target enumeration (the new version flattens vulns across all `Results` entries, which is fine for image scans but loses the per-target column).
> * The header docstring explaining the JSON→Python rationale (still useful context).
> 
> Not blocking, but worth knowing this is a behavioural change to the summary UX in addition to the gate change. If the team is OK with the simpler layout, fine.
> 
> **4. SBOM-as-artifact, not OCI-attached.** The CycloneDX SBOM is uploaded as a workflow artifact (90-day default retention). For supply-chain provenance, attaching the SBOM to the image via `cosign attach sbom` (or buildx provenance attestation) is the more durable pattern — survives artifact expiration and travels with the image. Out of scope for this PR but worth a follow-up.
> 
> **5. `scanners: vuln` on the SBOM step.** Technically harmless — CycloneDX is engaged by `format`, and `scanners: vuln` embeds vulnerability metadata in the SBOM. If that embedding is intentional (good for supply-chain), worth a comment saying so. If not, can drop.
> 
> **6. Earlier suppression commits (`99480ef ci: suppress 3 MEDIUM base image CVEs`) are superseded by the CRITICAL/HIGH-only gate.** Confirmed `.trivyignore` is empty in the final diff — good. Those MEDIUM CVEs will still appear in the Security tab via the all-severity SARIF scan, which is the right outcome.
> 
> **7. Trivy SARIF scan re-runs the full vulnerability analysis** (`severity: CRITICAL,HIGH,MEDIUM,LOW` vs hard-gate's `CRITICAL,HIGH`). Two scans of the same image is duplicate work, but Trivy's DB cache makes it cheap. Not worth optimising unless CI time becomes a concern.
> 
> ## Verified clean
> * `.trivyignore` correctly resolves via the root (main) checkout, not `build-src/.trivyignore` — closes the bypass vector noted in #173. Verified by re-reading the two-checkout layout in the workflow.
> * All Trivy action invocations pinned by SHA + version tag.
> * Push step has no `if: always()` — it skips on gate failure as expected.
> * Concurrency group correctly serialises by ref; push and dispatch converge on the same key when both target main.
> * `continue-on-error: true` on advisory steps doesn't mask the hard gate (which lacks it).
> * `if: always()` ordering means SARIF + SBOM + summary artifact upload all run on gate failure, so even blocked builds produce diagnostics.
> 
> ## Suggested action
> Decide intentionally on point 1 (trigger scope) and confirm point 2 (env-reviewer config in place), then this is good to merge. The Trivy gating itself is well-structured.

Re: Trigger doesn't match PR description — The push: branches: [main] trigger is intentional. The earlier description was written during the testing phase when we were exercising the workflow from a feature branch; the final approach is auto-CD on merge to main (same pattern as the existing Strata workflows). PR description updated to reflect this.

Re: AWS environment reviewers — Confirmed. The AWS environment has required_reviewers set to Zk2u and sapinb, so every deployment triggered by a main push still gates on an explicit approval before any ECR push happens. The auto-trigger doesn't bypass the protection.

Re: render_trivy_summary.py rewrite — Acknowledged. The severity-count table and MAX_FINDINGS truncation were intentionally simplified since the SARIF upload to the Security tab gives the full picture; the step summary is meant to be a quick pass/fail indicator rather than a full report. Happy to add the truncation guard back if the team prefers. SBOM-via-cosign noted as a follow-up item.

Good to merge from my side once you're satisfied on the above.

Zk2u:
Thanks — that resolves all three points cleanly.

Auto-CD-on-main + required-reviewers on the AWS env is the same posture I'd have wanted from a hand-rolled workflow_dispatch flow, just with the audit trail captured by the merge commit instead of a dispatch event. Good outcome.

On the summary rewrite: your reasoning makes sense — SARIF tab is the canonical full report, step summary as quick pass/fail indicator. I'd lean toward keeping a MAX_FINDINGS truncation guard purely as defensive UX (a sufficiently broken base image could otherwise render hundreds of rows into the summary page), but it's cheap-to-add-later and not blocking — happy to take it as-is.

LGTM, approving.

Zk2u (approved):
Approving — auto-CD-on-main is fine given AWS env required-reviewers are configured (Zk2u + sapinb), and the rest of the Trivy gating is well-structured. Truncation-guard nit is non-blocking; SBOM-via-cosign tracked as follow-up.

PR #223, opened 2026-05-08
cac/protocol: gate table-transfer receipts on local-transfer completion

Zk2u:
## Summary

Closes #217.

The garbler STF previously trusted `RecvTableTransferReceipt` as the sole authority for marking an evaluator's table as transferred and, once all receipts arrived, transitioning to `SetupComplete`. The `GarblingTableTransferred` action result for the local transfer was informational only.

Because the table commitments are public, an evaluator that knows them could send receipts for every slot immediately after the garbler entered `TransferringGarblingTables` and force `SetupComplete` without any actual transfer having occurred locally. That is a **soundness** break, not just a DoS — `SetupComplete` is supposed to mean "all evaluation tables transferred and persisted on this node."

## Changes

A per-slot `locally_transferred` flag is added to the `TransferringGarblingTables` step:

- Set by the `GarblingTableTransferred` action-completion handler when our worker reports a successful local transfer.
- Required by the `RecvTableTransferReceipt` handler before honoring a receipt; receipts whose slot has not been locally transferred are rejected.
- Also gates the `RecvTableTransferRequest` duplicate-suppression: a request that arrives after the local transfer is already complete is acked-and-ignored instead of dispatching another expensive `TransferGarblingTable` action.

The duplicate-request amplification when the first transfer is still in-flight is bounded upstream by per-peer protocol-message rate limiting (#220).

## Tests

Four new unit tests in `crates/cac/protocol/src/garbler/tests.rs`:

- `table_transfer_receipt_before_local_transfer_is_rejected` — soundness gate
- `table_transfer_receipt_succeeds_after_local_transfer_completes` — happy path
- `duplicate_table_transfer_request_after_local_transfer_emits_no_action` — amplification mitigation
- `garbling_table_transferred_action_result_marks_locally_transferred` — sanity for the gate's set side

## Issues
Closes #217.

## Validation
- `cargo check --workspace`
- `cargo clippy --workspace --tests --all-targets -- -D warnings`
- `cargo test -p mosaic-cac-protocol` (24 lib tests pass; 4 new garbler unit tests included)
- `cargo +nightly fmt --all --check`

## Notes
- Adds a field to a serialized state enum variant. Pre-release; no migration story required, but operators should be aware that existing persisted garbler state in `TransferringGarblingTables` will fail to deserialize after this change. (Setup-complete state is unaffected.)

Issue #217, opened 2026-05-03
Garbler STF accepts table transfer receipts as authority and re-enqueues duplicate transfer requests


Zk2u:
## Summary

Two related defects in the peer-driven garbling-table transfer flow introduced when transfer became request/receipt-keyed by table commitment:

1. **Duplicate-request amplification.** In the garbler STF's `TransferringGarblingTables` step, every valid `TableTransferRequest` for a known commitment emits a new `TransferGarblingTable` action. The code explicitly does not check whether that table is already requested, in-flight, or transferred. A configured malicious evaluator can send many duplicate requests for one valid commitment and enqueue repeated expensive garbling/bulk-transfer sessions.

2. **Receipt-as-authority bypass.** `RecvTableTransferReceipt` sets `transferred[pos] = true` and moves to `SetupComplete` as soon as all commitments have receipts. The `GarblingTableTransferred` action result no longer updates transfer state — it only logs. Therefore an evaluator that knows the expected commitments (which are public) can send receipts for all evaluation commitments immediately after the garbler enters `TransferringGarblingTables`, causing the garbler to persist `SetupComplete` even if no table transfer succeeded.

Previously, receipts were only accepted after the garbler observed all `TransferGarblingTable` completions and moved to `WaitForTableTransferReceipt`.

This is the reviewed outcome for Codex security finding [`6f0dc7af`](https://chatgpt.com/codex/cloud/security/findings/6f0dc7af47f481919affc5b78b6ae244).

## Impact

- (1) is a DoS amplification: each transfer opens a bulk stream and processes the full circuit, so duplicate requests can burn CPU, network, table-store I/O, and scheduler capacity.
- (2) is a **protocol soundness** issue: an evaluator can drive a garbler into `SetupComplete` without any successful table transfer. This invalidates the post-condition that `SetupComplete` ⇒ all evaluation tables transferred and persisted.
- Severity: high. (2) in particular is a soundness concern, not just availability.

## Proposed fix

Track transfer lifecycle explicitly and gate receipt acceptance on it.

Specifically:

- track requested / in-flight / successfully-transferred separately for each commitment in `TransferringGarblingTables`
- on `TableTransferRequest`:
  - if the commitment is already in-flight or already transferred, ack idempotently and **do not** emit a new `TransferGarblingTable` action
  - bound how many concurrent transfer actions can exist per peer
- on `GarblingTableTransferred` (action completion): mark the corresponding commitment as transferred and update STF state
[...]

PR #257, opened 2026-05-18
cac/protocol: defer evaluator SetupComplete until every receipt is acked

### LINKED ISSUE #256: Evaluator restore orphans in-flight SendTableTransferReceipt actions, deadlocking honest-crash recovery
## Summary

The evaluator state machine transitions `Step::ReceivingGarblingTables → Step::SetupComplete` in the same input handler that emits the final `SendTableTransferReceipt` tracked action(s). The `SetupComplete` restore branch does not re-emit `SendTableTransferReceipt`, so any receipt in flight at crash time is permanently orphaned. The garbler, waiting on those receipts in `Step::TransferringGarblingTables`, stalls indefinitely.

This is a liveness bug under **honest-crash recovery**, not a security one. fasm's contract is that restore must re-emit every tracked action that was emitted but not completed before the crash; this is the only mechanism the framework provides — nothing is replayed automatically. The whole safety story of the system rests on restore being correct, so this is critical.

## Mechanism

`crates/cac/protocol/src/evaluator/stf.rs:1002-1013` (in `handle_table_received`):

```rust
emit(
    actions,
    Action::SendTableTransferReceipt(TableTransferReceiptMsg::new(table_commitment)),
);

received[pos] = true;
debug!(...);

if received.all() {
    info!("all garbling tables received, setup complete");
    root_state.step = Step::SetupComplete;
}
```

And the completion handler is a deliberate no-op (`stf.rs:334-338`):

```rust
ActionResult::TableTransferReceiptAcked => {
    // The table transfer receipt message was sent. No further state change needed —
    // state was already advanced to SetupComplete when
    // we emitted the SendTableTransferReceipt action.
}
```

That comment is the bug's plain-language description: state advances to `SetupComplete` while the corresponding ack is still in flight.

Crash window: between persisting `Step::SetupComplete` and the framework processing `TrackedActionCompleted(SendTableTransferReceipt)`. Up to `N_EVAL_CIRCUITS` receipts can be in flight simultaneously because the SM executor batches completions through independent select branches.

On restart, `restore()` enters the `Step::SetupComplete` branch (`stf.rs:1096`) which only iterates deposits — `SendTableTransferReceipt` is never re-emitted.

## Consequence

- Garbler stays in `Step::TransferringGarblingTables` waiting on receipts that will never arrive.
- Evaluator considers itself `SetupComplete` and won't reattempt anything related to table transfers.
- No protocol signal exists for either side to detect this divergence — looks identical to a slow but healthy run.
- Plausible contributor to the Run 3 `no expectation for bulk transfer, dropping stream` symptom Madan reported in the external-operator gap report (G-7 territory): evaluator restart past `SetupComplete`, garbler retries transfer for a slot the evaluator no longer expects.

While `Step::ReceivingGarblingTables` (`restore()` at `stf.rs:1078-1094`) does re-emit `SendTableTransferReceipt` for every `received[pos] = true`, that recovery path is only reachable while the step is still `ReceivingGarblingTables`. The moment `received.all()` triggers the transition, it stops being reachable for the receipts that triggered the transition.

## Fix

Mirror the pattern from the now-closed #223 without the soundness framing — purely as restore correctness:

Add `receipt_acked: HeapArray<bool, N_EVAL_CIRCUITS>` to `Step::ReceivingGarblingTables`:

1. `handle_table_received` no longer transitions on `received.all()`; just sets `received[pos] = true` and emits the receipt.
2. `TableTransferReceiptAcked` handler matches on `ActionId::SendTableTransferReceipt(commitment)`, finds the slot, sets `receipt_acked[pos] = true`. If `received.all() && receipt_acked.all()`, transitions to `SetupComplete`.
3. `restore()` for `ReceivingGarblingTables` gates the receipt re-emit on `received[i] && !receipt_acked[i]` (the current code re-emits for every `received[i]`, which is also over-eager — re-sends receipts that already acked on every restart).

This is the same invariant the garbler side already maintains for its own emit-and-transition transitions (advance the step only after all dependent action completions are observed).

## Broader audit context

I walked every state transition in both garbler and evaluator state machines that does \`emit + transition step\` in the same handler, checking whether the destination step's restore re-emits the just-emitted actions. The pattern is safe everywhere else:

- Garbler `SendingCommit → WaitingForChallenge` — transition on ack, no emit at transition ✓
- Garbler `SendingChallengeResponse → TransferringGarblingTables` — no emit at transition ✓
- Evaluator `WaitingForCommit → WaitingForChallengeResponse` — emit `SendChallengeMsg`, restore re-emits ✓
- Evaluator `WaitingForChallengeResponse → VerifyingOpenedInputShares` — emit `VerifyOpenedInputShares`, restore re-emits ✓
- Evaluator `VerifyingOpenedInputShares → VerifyingTableCommitments` — emit `GenerateTableCommitment` per circuit, restore re-emits per `!verified` ✓
- Evaluator `VerifyingTableCommitments → ReceivingGarblingTables` — emit `ReceiveGarblingTable` per slot, restore re-emits per `!received` ✓
- Evaluator deposit `GeneratingAdaptors → SendingAdaptors` — emit `DepositSendAdaptorMsgChunk` per chunk, restore re-emits per `!acked` ✓
- **Evaluator `ReceivingGarblingTables → SetupComplete` — emit `SendTableTransferReceipt`, restore for `SetupComplete` does NOT re-emit it ✗** (this issue)

So just the one gap.

cc @sapinb @AaronFeickert

Zk2u:
## Description

The evaluator transitioned `Step::ReceivingGarblingTables → Step::SetupComplete` in the same handler that emitted the final `SendTableTransferReceipt` tracked action(s). Up to `N_EVAL_CIRCUITS` receipts could be in flight simultaneously when the transition fired. `restore()` for `SetupComplete` does not (and cannot) re-emit `SendTableTransferReceipt`, so a crash between the persisted step transition and `TrackedActionCompleted` delivery permanently orphaned the unacked receipts — leaving the garbler stalled in `TransferringGarblingTables` indefinitely.

fasm's contract is that `restore` re-emits every tracked action whose completion the framework has not yet processed; the framework replays nothing automatically. The data needed to do so must remain in the step until the ack lands.

### Fix

Add `receipt_acked: HeapArray<bool, N_EVAL_CIRCUITS>` to `ReceivingGarblingTables`. `handle_table_received` no longer transitions on `received.all()` alone. `TableTransferReceiptAcked` matches on `ActionId::SendTableTransferReceipt(commitment)`, locates the slot in `eval_commitments`, sets `receipt_acked[pos] = true`, and only transitions to `SetupComplete` when `received.all() && receipt_acked.all()`. `restore()` gates the receipt re-emit on `!receipt_acked[i]` (the previous code re-emitted for every `received[i]`, which was also over-eager — it re-sent already-acked receipts on every restart).

### Type of Change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Refactor
- [x] New or updated tests
- [ ] Dependency update
- [ ] Security fix

Breaking: the `Step::ReceivingGarblingTables` variant adds a new `receipt_acked` field. Persisted evaluator states from before this change in that step will fail to deserialize (postcard `DeserializeUnexpectedEnd`). Pre-testnet, state reset is the recovery path; a versioned-storage upgrade binary is planned separately.

## Notes to Reviewers

This is a **liveness bug under honest crash**, not security — distinct from the #217 / #223 discussion which was about adversarial receipt behaviour. The system's safety story rests on "restart is safe", and every tracked action in flight must be reachable to `restore()` so the framework can drive it to completion.

Audit context: I walked every `emit + transition step` site in both state machines. This is the only gap of its kind — every other transition either emits no action at the transition or transitions into a step whose `restore` re-emits the just-emitted actions.

Tests added:

- `handle_table_received_does_not_advance_to_setup_complete` — invariant: receipt emitted, step stays `ReceivingGarblingTables`.
- `table_transfer_receipt_acked_advances_only_when_all_received_and_acked` — full ack sequence; `SetupComplete` fires on final ack.
- `duplicate_receipt_ack_is_idempotent` — restore-emitted receipts can produce a second ack; must not error.
- `late_receipt_ack_after_setup_complete_is_ignored` — late ack past the transition is a no-op.
- `restore_receiving_tables_re_emits_unacked_receipts_and_pending_receives` — mixed-state restore covers all three slot kinds.
- `restore_receiving_tables_emits_nothing_when_all_received_and_acked` — steady-state restore must emit nothing.
- `ack_with_unknown_commitment_is_invalid_input` — corrupted action id fails closed.

Existing `cac/protocol` e2e test (`test_e2e`) was adjusted: the step is now `ReceivingGarblingTables` (not `SetupComplete`) between `GarblingTableReceived` and the receipt acks.

## Checklist

## Related Issues

closes #256

### DIFF
--- a/crates/cac/protocol/src/evaluator/stf.rs
+++ b/crates/cac/protocol/src/evaluator/stf.rs
@@ -1137,15 +1137,25 @@ pub(crate) async fn restore<S: StateRead>(
             receipt_acked,
             ..
         } => {
-            for ii in 0..eval_commitments.len() {
-                let commitment = eval_commitments[ii];
-                if !received[ii] {
+            for i in 0..eval_commitments.len() {
+                let commitment = eval_commitments[i];
+                // (!received && receipt_acked) is unreachable by construction:
+                // `receipt_acked[i]` can only be set inside this step, and the
+                // ack handler requires `received[i] = true` to have been set
+                // previously (the receipt was emitted in
+                // `handle_table_received`). Asserting the invariant here keeps
+                // the loop honest if future changes accidentally violate it.
+                debug_assert!(
+                    !receipt_acked[i] || received[i],
+                    "receipt_acked[{i}] must imply received[{i}]"
+                );
+                if !received[i] {
                     // Combined action: registers expectation, sends request,
                     // receives table.
                     emit(actions, Action::ReceiveGarblingTable(commitment));
                     continue;
                 }
-                if !receipt_acked[ii] {
+                if !receipt_acked[i] {
                     // Receipt for this slot was emitted in `handle_table_received`
                     // but not yet acked by the network layer. Re-emit so the
                     // garbler is guaranteed to learn this table was received.
--- a/crates/cac/protocol/src/evaluator/stf.rs
+++ b/crates/cac/protocol/src/evaluator/stf.rs
@@ -1047,6 +1047,18 @@ async fn handle_table_received<S: StateMut>(
                 return Err(SMError::InvalidInputData);
             };
 
+            // Validate the commitment first so a mismatched duplicate still
+            // aborts (matching the original behaviour) instead of being
+            // silently dropped by the idempotency guard below.
+            let expected_commitment = eval_commitments[pos];
+            if table_commitment != expected_commitment {
+                warn!(%index, "evaluator received garbling table with mismatched commitment, aborting");
+                root_state.step = Step::Aborted {
+                    reason: format!("invalid table for index {}", index),
+                };
+                return Ok(());
+            }
+
             // Idempotency: a duplicate `GarblingTableReceived` for an already-
             // marked slot would otherwise emit a stray `SendTableTransferReceipt`.
             // The framework should not deliver completions twice, but the
@@ -1054,15 +1066,6 @@ async fn handle_table_received<S: StateMut>(
             // guard in `TableTransferReceiptAcked`.
             if received[pos] {
                 debug!(%index, "duplicate GarblingTableReceived; ignoring");
-                return Ok(());
-            }
-
-            let expected_commitment = eval_commitments[pos];
-            if table_commitment != expected_commitment {
-                warn!(%index, "evaluator received garbling table with mismatched commitment, aborting");
-                root_state.step = Step::Aborted {
-                    reason: format!("invalid table for index {}", index),
-                };
                 return Ok(());
             }
 
@@ -1254,6 +1257,13 @@ pub(crate) async fn restore<S: StateRead>(
             }
         }
         Step::SetupConsumed { .. } => {}
+        // NOTE: an abort transition from `ReceivingGarblingTables` (commitment
+        // mismatch in `handle_table_received`) can leave earlier slots'
+        // `SendTableTransferReceipt` actions in flight, which this branch
+        // does not re-emit. Functionally acceptable here because the SM is
+        // terminating and will not act on the result, but it does leave the
+        // garbler peer without those receipts. Cleaner abort-propagation is
+        // tracked as a follow-up (see PR #257 review thread).
         Step::Aborted { .. } => {}
     }
 
--- a/crates/cac/protocol/src/evaluator/stf.rs
+++ b/crates/cac/protocol/src/evaluator/stf.rs
@@ -384,6 +384,10 @@ pub(crate) async fn handle_action_result<S: StateMut>(
                 | Step::Aborted { .. } => {
                     debug!("late table transfer receipt ack after step advanced; ignoring");
                 }
+                // `SendTableTransferReceipt` is only emitted from
+                // `ReceivingGarblingTables`, and the step never moves
+                // backward. An ack landing here would imply a framework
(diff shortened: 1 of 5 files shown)


Zk2u:
Independent review round 2 surfaced one adjacent concern worth tracking as a follow-up (not in this PR's scope):

**Abort-during-ReceivingGarblingTables can orphan in-flight `SendTableTransferReceipt` actions across crash.** `handle_table_received` transitions to `Aborted` on commitment mismatch, but receipts emitted for earlier valid slots remain in flight. `restore()` for `Aborted` emits nothing.

This is qualitatively different from the bug closed here: the local SM is aborted and won't progress regardless, so the orphan doesn't cause a wedged evaluator. The downstream concern is garbler-side uncertainty — the garbler doesn't get an abort signal and may keep waiting for the missing receipts. That's an abort-propagation gap, not a restore-correctness gap. Filing as a follow-up rather than expanding scope here.

sapinb (approved):
ACK ccc513e93163ecf762a7f0d51a7bd71285aaedf2

PR #259, opened 2026-05-22
test: don't hardcode reserved index in tests

AaronFeickert:
## Description

Recent work in #229 improves validation for challenge indexes. However, two of its added tests unnecessarily hardcode the reserved index. This PR moves to built-in functionality to dynamically pull the reserved index as a hedge against future changes to this value. It also removes the silent use of a default index value in tests, and fixes an errant comment.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [x] New or updated tests
- [ ] Dependency update
- [ ] Security fix

## Notes to Reviewers

N/A

## Checklist

## Related Issues

Follows up on #229.


AaronFeickert:
@Zk2u also note that the test functionality introduced in #229 claims to handle the case of out-of-bounds indexes. This is not the case (aside from the reserved index), and this is never tested: it is not possible to fully override the `Index` constructor due to the index being a private field.

sapinb (approved):

PR #210, opened 2026-05-02
Fix: verify opened output shares

nakkstar123:
## Description

Updates evaluator stf to verify opened output shares, fixing a bug caught by audit. 

### Type of Change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature/Enhancement (non-breaking change which adds functionality or enhances an existing one)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Refactor
- [x] New or updated tests
- [ ] Dependency update
- [ ] Security fix

## Notes to Reviewers

The suggested fix in v12's audit is correct, and that's what this PR implements. Additionally, I've added tests that validate correct behavior. In particular, evaluator now aborts if an output share was corrupted (and does not store the shares), which wasn't happening earlier. 

## Checklist

- [x] I have performed a self-review of my code.
- [x] I have commented my code where necessary.
- [ ] I have updated the documentation if needed.
- [x] My changes do not introduce new warnings.
- [x] I have added tests that prove my changes are effective or that my feature works.
- [x] New and existing tests pass with my changes.

## Related Issues

Fixes #195.

PR #264, opened 2026-05-27
Fix: verify opened input share indices

### LINKED ISSUE #233: Verify opened share indices match challenge indices
### Describe the bug

After challenge, evaluator accepts shares from the garbler and verifies them against polynomial commitments using the share's self-reported index, without binding that index to the position the protocol actually challenged. This was recently caught for the output shares in #210, but the issue exists for input shares as well.

Some context: `Share` is a `(Index, Scalar)`. Both single-share and batch verification use the share's embedded index as the evaluation x-coordinate. Either function will accept any valid `(j, P(j))` pair on the polynomial, including j values that were never challenged. The cut-and-choose protocol's soundness depends on the verifier binding each opened share to the challenged index for that position; otherwise, a malicious garbler can substitute honest shares from positions of their choice and avoid revealing positions where they cheated.

Specifically, these are two places that need fixing:

1. Evaluator stf validates that `response_msg_chunk.circuit_index` is in the challenge set, then stores the chunk's shares under that key. Nothing checks that the `Shares` inside the chunk have `.index().get() == circuit_index`. Batch verification later trusts the embedded index, so a garbler can put `(j, P(j))` for honest j ≠ i into a chunk claimed for challenge index i and pass verification. Fix: Before `put_opened_input_shares_chunk`, assert every share in `response_msg_chunk.shares` has `.index().get() == response_msg_chunk.circuit_index as usize` and abort on mismatch.
2.  Evaluator stf's`verify_reserved_setup_input_shares`compares `g^{P(0)}` against `g^{reserved_share.value()}`, never reading `reserved_share.index()`. The consumer uses the stored share as-is (`reserved_setup_shares.to_vec()`) before feeding it to Lagrange interpolation, which treats `share.index()` as the x-coordinate. A wrong index gives incorrect interpolation and breaks evaluation. The deposit/withdrawal paths already handle this defensively with Share::new(Index::reserved(), share_value). Fix: Add `reserved_share.index() == Index::reserved()`to `verify_reserved_setup_input_shares`.

### Steps to reproduce

N/A

### Platform(s)

_No response_

### Code of Conduct

- [x] I agree to follow the Code of Conduct

nakkstar123:
## Description

Addresses #233 by enforcing correct share indices for input shares as well (output shares addressed in #210). Adds corresponding tests that verify the fix. 

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [x] New or updated tests
- [ ] Dependency update
- [ ] Security fix

## Notes to Reviewers

Used Claude Code for writing tests. They failed before the change and pass now, as expected. 

## Checklist

## Related Issues

Fixes #233.

### DIFF
--- a/crates/cac/protocol/src/evaluator/stf.rs
+++ b/crates/cac/protocol/src/evaluator/stf.rs
@@ -910,9 +910,15 @@ async fn handle_recv_challenge_response_msg<S: StateMut>(
                         warn!(
                             circuit_index = %expected_index,
                             share_index = %share.index(),
-                            "evaluator opened input share index mismatch, rejecting chunk"
+                            "evaluator opened input share index mismatch, aborting"
                         );
-                        return Err(SMError::InvalidInputData);
+                        root_state.step = Step::Aborted {
+                            reason: format!(
+                                "invalid opened input share: chunk circuit_index {expected_index} carries share with index {}",
+                                share.index()
+                            ),
+                        };
+                        return Ok(());
                     }
                 }
             }
--- a/crates/cac/protocol/src/evaluator/stf.rs
+++ b/crates/cac/protocol/src/evaluator/stf.rs
@@ -897,6 +897,26 @@ async fn handle_recv_challenge_response_msg<S: StateMut>(
                 return Err(SMError::InvalidInputData);
             }
 
+            // Bind every embedded share index to the chunk's circuit_index.
+            // batch_verify_shares uses share.index() as the evaluation
+            // x-coordinate, so without this a garbler could substitute
+            // honest (j, P(j)) shares into a chunk claimed for circuit i and
+            // pass verification — defeating cut-and-choose soundness.
+            let expected_index = Index::new(response_msg_chunk.circuit_index as usize)
+                .expect("circuit_index validated against challenge set above");
+            for wire_shares in response_msg_chunk.shares.iter() {
+                for share in wire_shares.iter() {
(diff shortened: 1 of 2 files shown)


crates/cac/protocol/src/evaluator/stf.rs
```diff
@@ -880,6 +897,26 @@ async fn handle_recv_challenge_response_msg<S: StateMut>(
...
+                .expect("circuit_index validated against challenge set above");
+            for wire_shares in response_msg_chunk.shares.iter() {
+                for share in wire_shares.iter() {
+                    if share.index() != expected_index {
+                        warn!(
+                            circuit_index = %expected_index,
+                            share_index = %share.index(),
+                            "evaluator opened input share index mismatch, rejecting chunk"
+                        );
+                        return Err(SMError::InvalidInputData);
```
sapinb: Output share index mismatch and Reserved share index mismatch lead to a protocol Abort.
But Input share index mismatch is an invalid input err, that allows retries.

This should be made consistent.

AaronFeickert:
This supersedes #210 since it includes those commits as well; correct?

Zk2u:
Doesn't supersede, just stacks on top.

AaronFeickert (approved):

sapinb (approved):
ACK 9d5c01e8c5cbbdf22bd49cbe87355c8c4d3147a1

PR #273, opened 2026-06-07
fix: stop shipping reduced-circuits test binary in production Docker image

### LINKED ISSUE #212: Production Docker image ships reduced-circuits binary, selectable via env var

mvanhorn:
## Description

The production Docker image no longer ships or selects the reduced-circuits test binary. The Dockerfile previously built mosaic with the reduced-circuits feature alongside the production binary and the entrypoint chose between them via `MOSAIC_REDUCED_CIRCUITS`, so a single env var flipped a production deployment onto test circuits. The image now builds only the production binary and the entrypoint always execs it, keeping the existing config-file presence check.

Closes #212

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency update
- [x] Security fix

## Notes to Reviewers

The diff is removal-heavy on purpose: the reduced-circuits build stage and the entrypoint branch both go away, and the entrypoint keeps its config check then execs `/usr/local/bin/mosaic` unconditionally. `bash -n docker/entrypoint.sh` passes; anyone needing reduced circuits for local testing still has the cargo feature, it just no longer rides inside the production image.

## Checklist

- [x] I have added tests that prove my changes are effective or that my feature works. (N/A: Docker packaging change; no test harness covers image contents. Verified with `bash -n docker/entrypoint.sh` and by walking the remaining build stages.)

## Related Issues

Closes #212

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -32,12 +32,9 @@ COPY .cargo .cargo
 COPY bin bin
 COPY crates crates
 
-# Build both release binaries so runtime can switch via env var.
+# Build the production release binary.
 RUN cargo build --release --bin mosaic \
-    && cp target/release/mosaic target/release/mosaic-default \
-    && cargo build --release --features=reduced-circuits --bin mosaic \
-    && mv target/release/mosaic target/release/mosaic-unsafe-test \
-    && strip --strip-debug target/release/mosaic-default target/release/mosaic-unsafe-test
+    && strip --strip-debug target/release/mosaic
 
 # ------------------------------------------------------------------------------
 # Stage 2: Minimal runtime image
@@ -65,8 +62,7 @@ RUN FDB_ARCH=$([ "$TARGETARCH" = "arm64" ] && echo "aarch64" || echo "amd64") \
 #   FDB cluster file                   – path configured via storage.cluster_file in config.toml
 RUN mkdir -p /etc/mosaic && chown ubuntu:ubuntu /etc/mosaic
 
-COPY --from=builder /build/target/release/mosaic-default /usr/local/bin/mosaic
-COPY --from=builder /build/target/release/mosaic-unsafe-test /usr/local/bin/mosaic-unsafe-test
+COPY --from=builder /build/target/release/mosaic /usr/local/bin/mosaic
 COPY docker/entrypoint.sh /usr/local/bin/entrypoint.sh
 RUN chmod +x /usr/local/bin/entrypoint.sh
 
--- a/docker/entrypoint.sh
+++ b/docker/entrypoint.sh
@@ -8,20 +8,6 @@
 # The config file's [circuit] section should reference:
 #   path = "/etc/mosaic/circuit.v5c"
 #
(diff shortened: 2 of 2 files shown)


sapinb (commented):
We currently use reduced circuit count builds for internal test deployments.
Also seeing high memory usage during concurrent setups with real-sized circuit and full circuit counts.
Should hold off merging this until we have more memory optimizations in and transition to using the full circuit counts for deployments.

mvanhorn:
Understood, and that is a good reason to wait. Leaving it open as-is just parks a PR in your review queue that nobody should merge yet, so pick whichever suits you.

I can convert this to a draft, which keeps the diff and the discussion but takes it out of the ready-for-review list until you transition to full circuit counts. Ping me here when that lands and I will mark it ready and rebase.

Or if the transition is far enough out that an open PR is just noise, I will close it. The change is small, four lines out of the Dockerfile plus the entrypoint cleanup, so recreating it later costs nothing.

One thing worth checking whenever you do pick it up: this removes the reduced-circuits binary from the production image only. If your internal test deployments pull that same image tag rather than building their own, they would lose the binary too, and the right shape would be a build arg that keeps it for the test image. Tell me how your test deployments get built and I will adjust it if that is the case.

PR #275, opened 2026-06-08
feat(rpc): load circuit metadata from v5c header

### LINKED ISSUE #266: Add RPC call for circuit checksum
It would be useful to allow the bridge node to query Mosaic for its active circuit. This is especially relevant since this encompasses the SP1 verification key.

Since Mosaic has no knowledge itself of the verification key, it should suffice to have an RPC call for the circuit checksum, which is already known. The bridge node is then responsible for maintaining a mapping between this checksum and any SP1-specific data it must track.

sapinb:
## Description
Loads circuit metadata from header of configured v5c circuit and serves on `mosaic_getCircuitDefs` rpc endpoint.

This is intended to be used by bridge to verify that mosaic is using the expected circuit corresponding to the counterproof VK that it is using.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency update
- [ ] Security fix

## Notes to Reviewers
The rpc endpoint already existed but used default values. This PR populates it with correct metadata.

## Checklist

## Related Issues

fixes #266

### DIFF
--- a/crates/rpc/types/src/circuit.rs
+++ b/crates/rpc/types/src/circuit.rs
@@ -1,3 +1,9 @@
+use std::{
+    io::{self, Read},
+    path::Path,
+};
+
+use ckt_fmtv5_types::v5::c::{HEADER_SIZE, HeaderV5c};
 use serde::{Deserialize, Serialize};
 
 use crate::RpcByte32;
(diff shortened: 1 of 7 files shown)


crates/rpc/types/src/circuit.rs
```diff
@@ -14,20 +20,33 @@ pub struct RpcCircuitInfoEntry {
...
-            name: "default".into(),
-            commitment: [0; 32].into(),
+    /// Create circuit info by reading the v5c header from the circuit file at `path`.
+    pub fn from_circuit_file(path: &Path) -> io::Result<Self> {
+        let mut f = std::fs::File::open(path)?;
+        let mut header_bytes = [0u8; HEADER_SIZE];
+        f.read_exact(&mut header_bytes)?;
+        let header = HeaderV5c::from_bytes(&header_bytes)?;
+
+        let file_size = std::fs::metadata(path)?.len();
```
AaronFeickert, line 30: We don't check that this size correlates with the gate count. Is this fine?

crates/rpc/types/src/circuit.rs
```diff
@@ -14,20 +20,33 @@ pub struct RpcCircuitInfoEntry {
...
+
+        let name = path
+            .file_stem()
+            .and_then(|s| s.to_str())
+            .unwrap_or("unknown")
```
AaronFeickert, line 35: Another approach is to make `name` into an `Option` for clarity.

AaronFeickert: Or we could set the unknown name to an empty string.

AaronFeickert (approved):

PR #284, opened 2026-06-14
fix: harden SM-executor ack & bulk paths against backpressure stalls

Zk2u:
Replaces #282 (closed when the branch was renamed; same commits).

---

## Summary

Fixes the ~30s SM-executor freeze during testnet (4-operator cascade from Madan's report).

`request.ack().await` was called inline from the executor's select loop. `Stream::write` awaits `recv_buffer()`, which blocks until the peer drains. If one peer's QUIC receive window filled, the ack call blocked the executor for every other peer too.

Four changes:

1. **Spawn the ack** (`crates/sm-executor/src/lib.rs`). `handle_inbound_request` now `monoio::spawn`s the ack with a 5s timeout instead of awaiting inline. Acks are best-effort — if it doesn't land the peer retransmits.
2. **Cap inflight acks at 1024** via `tokio::sync::Semaphore::try_acquire_owned`. Saturation drops with a warn log instead of spawning more tasks.
3. **30s per-chunk timeout on bulk translation send** (`crates/job/executors/src/garbler.rs`). Same backpressure problem on the 2 MiB chunks; timeout surfaces as `CircuitError::TransientFailure` so the scheduler retries. Uses a local `futures_timer::Delay`-based helper because the trait future needs to be `Send` and `monoio::time::timeout` isn't.
4. **Wrap `NetClient::send`'s write in `ack_timeout`** (`crates/net/client/src/lib.rs`). Previously only the ack-read had a timeout; the write could hang on backpressure.

Removed `SmExecutorError::Ack` since acks no longer return errors to the loop.

## Test plan

- [x] `cargo test --release -p mosaic-sm-executor -p mosaic-net-client -p mosaic-job-executors` passes
- [x] `cargo test --release -p mosaic-cac-protocol` passes (44/44, e2e in 16.6s)
- [x] `cargo clippy --workspace --all-targets -- -D warnings` clean
- [x] `cargo fmt --all` clean
- [ ] Cluster: confirm a stalled peer no longer cascades
- [ ] Cluster: heavy concurrent inbound load doesn't backlog acks

## Known gaps (follow-ups, not blocking)

- No unit test for the new ack-spawn happy path, timeout, or semaphore saturation. Hitting these needs a real STF success setup — non-trivial scaffolding. Behavior is covered by the existing testnet repro and the `inbound_stf_failure_does_not_send_ack` test.
- No metric for ack saturation or background-ack timeouts (only `tracing::warn!`).


crates/sm-executor/src/lib.rs
```diff
@@ -209,6 +217,8 @@ where
     job_handle: JobSchedulerHandle,
     net_client: NetClient,
     command_rx: kanal::AsyncReceiver<SmCommand>,
+    /// Bounds the number of in-flight background acks (see [`MAX_INFLIGHT_ACKS`]).
+    inflight_acks: Arc<tokio::sync::Semaphore>,
```
sapinb: Can this bound be per-peer ? With a single global limit, it is still possible for a single misbehaving peer to degrade, or in the worst case halt, the protocol communication of the other peers between themselves.

crates/sm-executor/src/lib.rs
```diff
@@ -800,15 +811,62 @@ where
...
+                    }
+                    Err(_) => {
+                        tracing::warn!(
+                            peer = ?peer_id,
+                            timeout = ?ACK_BACKGROUND_TIMEOUT,
+                            "background ack timed out; peer not draining stream",
+                        );
+                    }
+                }
+                drop(permit);
```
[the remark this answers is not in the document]
Zk2u: Removed the semaphore entirely in 1aaffff. The cap was trying to bound stuck QUIC writes but couldn't (permit released on timeout-drop while write_task stayed blocked). It also wasn't needed: net-svc's `MAX_CONCURRENT_BIDI_STREAMS = 100` per connection already caps per-peer stream concurrency, and a peer can't have more spawned ack tasks than open streams to us.

crates/sm-executor/src/lib.rs
Zk2u: Removed the semaphore in 1aaffff. The per-peer concern resolves because net-svc's `MAX_CONCURRENT_BIDI_STREAMS = 100` per connection is already per-peer — a misbehaving peer can hold at most 100 spawned ack tasks to us regardless of what other peers are doing.

sapinb (approved):
ACK 1aaffffaacaa603b9bd8d9ebfd5b96632636088b

PR #289, opened 2026-06-22
job/executors: heartbeat progress logs for garbling phases

Zk2u:
## Summary

Adds periodic INFO heartbeat logs during long-running garbling phases (table upload G8, table eval E8, commitment generate G3/E3) so operators can tell at a glance whether a node is making progress or wedged.

## Approach

- New `HeartbeatTracker` helper in `crates/job/executors/src/progress.rs` with `Bytes` / `Blocks` units and a 30s default period.
- No spawned timer — sessions tick `maybe_log` per chunk, the tracker checks elapsed time and only emits when the period has passed. Cheap on the not-yet-time path.
- Wired into three `CircuitSession` impls:
  - **`TransferSession`** (G8) → `table.upload` (bytes, total = `header.and_gates × 16`)
  - **`CommitmentSession`** (G3 / E3) → `garbling.commit` / `garbling.verify` (gates, total = `header.total_gates()`)
  - **`EvaluationSession`** (E8) → `table.eval` (gates, total = `header.total_gates()`)

## Log format

Emitted on tracing target `mosaic_progress`. Operators filter via `RUST_LOG=mosaic_progress=info` or grep `mosaic_progress` in their log pipeline.

```
table.upload progress id=0xabcdef01 done=<X>/<Y> pct=<Z>% inst=<>/s avg=<>/s elapsed=<> eta=<>
table.eval progress id=0xabcdef01 pct=~<Z>% elapsed=<>
garbling.commit progress id=ckt<N> pct=~<Z>% elapsed=<>
```

`id` is the first 4 bytes of the commitment hash for table phases, or the circuit index for commitment-generate phases. Bytes show throughput + ETA when total is known; block-counted phases show percent + elapsed only (no rate / ETA — gate processing isn't byte-paced).

## Test plan

- [x] Unit tests on the helper: byte/duration/rate formatting, period gating, emit-after-period-elapsed.
- [x] `cargo test --release -p mosaic-job-executors` — 15 / 15 pass.
- [x] `cargo +nightly fmt --all`, `cargo clippy -p mosaic-job-executors --all-targets -- -D warnings` — clean.
- [ ] Manual e2e: observe heartbeats during a real garbling phase, confirm cadence and totals.

## Draft

Open as a draft for review of cadence (30s) + log format + `id=` field choices before promoting.

## Out of scope

- Heartbeat for the evaluator-side bulk receive (E4) — separate pool action surface.
- Metrics export — same data points should later feed a Prometheus surface.
- Configurable period — hardcoded const for v1.

### DIFF
--- a/crates/job/executors/src/circuit_sessions.rs
+++ b/crates/job/executors/src/circuit_sessions.rs
@@ -44,7 +44,20 @@ use mosaic_net_client::BulkSender;
 use mosaic_storage_api::table_store::TableReader;
 use mosaic_vs3::{Index, Scalar, Share};
 
-use crate::garbling::{GarblingSession, GarblingSetup, compute_commitment, hash_garbling_params};
+use crate::{
+    garbling::{GarblingSession, GarblingSetup, compute_commitment, hash_garbling_params},
+    progress::{HEARTBEAT_PERIOD, HeartbeatTracker, ProgressUnit},
+};
+
+/// Short-id helper used to label progress logs. Renders the first 4 bytes
+/// of a digest as `0x........`. Truncates gracefully if shorter.
+fn short_id(digest: &[u8]) -> String {
+    let mut s = String::from("0x");
+    for b in digest.iter().take(4) {
+        s.push_str(&format!("{b:02x}"));
+    }
+    s
+}
 
 // ════════════════════════════════════════════════════════════════════════════
 // Gate parsing helpers for OwnedBlock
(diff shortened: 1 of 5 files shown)


sapinb:
"table.upload" heartbeat can be stalled by network backpressure in garbler -> evaluator connection or evaluator s3 upload.
In the worst case, this can delay heartbeat by upto `concurrency_per_garbler_worker` * `chunk_timeout`.
If a garbling batch contains such delayed "table.upload" and other non network dependent jobs, their heartbeats will also be similarly delayed.
Not a blocker, but operators will have to consider that heartbeat logs may fluctuate between `HEARTBEAT_PERIOD` and `concurrency_per_garbler_worker` * `chunk_timeout`.

Additionally, `session timed out on chunk — evicting for retry` or `session error — evicting for retry` are other useful related logs to track that can replace the heartbeat as the final log for the job if it gets dropped pending retry. 

Note: `concurrency_per_garbler_worker` = `job_scheduler.garbling.max_concurrent` / `job_scheduler.garbling.worker_threads`.

@alexhui01 @madan-oss

sapinb (approved):
ACK dfe42fa6570dc31141da5fa2471faafd3b5e85cf

Issue #291, opened 2026-06-24
One unresolvable peer at startup blocks mosaic from booting and from peering with all healthy peers


madan-oss:
### Describe the bug

At startup mosaic resolves **every** `[[network.peers]]` address up front and
treats a single unresolvable hostname as fatal — the process exits:

```
Error: failed to resolve address `testnet-mosaic-six.chainflow.xyz:7000`
Caused by: failed to lookup address information: Name or service not known
```

This is an **all-or-nothing** gate. Because the process exits, it never starts —
and therefore **never peers with the peers that are already up and reachable**.
One offline operator takes down connectivity to every online operator.

Observed on **Testnet Alpha (testnet-prod)** during an uncoordinated rollout:
mosaic crash-loops (CrashLoopBackOff) on all 3 Alpen operators because the
Stakely and Chainflow mosaic P2P hostnames don't have DNS yet. This also cascades
into `bridge-client`, which can't reach its local mosaic and panics (tracked
separately on `strata-bridge`).

- **Commit/image:** `mosaic:376d1bef`
- **Severity:** High — breaks uncoordinated/rolling (staggered) deployments
- **Crate:** peer resolution / startup path (`[[network.peers]]`)

### Steps to reproduce

1. Configure N peers where ≥1 hostname has no DNS yet and the rest are healthy/resolvable.
2. Start mosaic.
3. Observe the process exits non-zero on the first unresolvable peer → CrashLoopBackOff.
4. Note the reachable peers are never dialed — the node contributes nothing to the mesh.

### Impact

- **Blocks own startup** — the daemon can't run while any single peer name is unresolvable.
- **Blocks peering with healthy peers** — since it never boots, online peers get zero
  connectivity from this node; the mesh can't begin partial assembly and is gated on the
  last/slowest operator.
- Contradicts the "uncoordinated, mesh self-assembles" model and creates a chicken-and-egg
  dependency: a node can't boot until all peers publish DNS → which needs their LBs → which
  needs them deployed first.

### Expected behavior

Unresolvable/unreachable peers should be **non-fatal and independent**: boot on
self/storage only, peer with whoever resolves now, and lazily resolve + retry the
rest with backoff.

**Suggested fix:** move peer resolution out of the fatal startup path into the
per-peer connect/retry loop (resolve-on-dial, per-peer isolation).

### Platform(s)

Linux (x86)

sapinb:
resolved in #292

PR #292, opened 2026-06-24
net/svc,bin/mosaic: resolve peer DNS lazily, don't crash on NXDOMAIN

Zk2u:
## Summary

Fixes a testnet startup crash: mosaic exited at boot if any peer's hostname returned NXDOMAIN, so every operator's node crash-looped until the slowest external operator published their DNS A record. Uncoordinated rollouts couldn't make progress.

Switches peer addresses from eagerly-resolved `SocketAddr` to a host+port pair that's re-resolved on every outbound connection attempt. DNS coming online (or being rotated) is picked up naturally without coordinated restarts.

## Behavior changes

- **Startup**: `NetService::new` no longer requires every peer hostname to resolve. Unresolved peers log a warn and are skipped from the pre-handshake peer-id prediction map (TLS still authenticates inbound regardless). Startup-ready signal is held until best-effort DNS lookups complete, so callers don't see `Ok` return before the service can actually process commands. Lookups run in parallel with a 2s per-peer cap.
- **Outbound attempts**: each attempt re-resolves the hostname before connecting. NXDOMAIN at connect time fails the attempt and falls through to the normal reconnect-backoff schedule. DNS budget is `CONNECTION_TIMEOUT / 2` (2.5s), QUIC connect budget is `CONNECTION_TIMEOUT` (5s); errors are phase-tagged for triage.
- **IP rotation**: an operator changing their A record is picked up on the next outbound attempt without restart.

## Config surface

`PeerConfig::addr: SocketAddr` is replaced with `host: String, port: u16`. The convenience constructor `new(peer_id, SocketAddr)` is kept so tests using IP literals didn't need to change. New `from_host(peer_id, host, port)` is the production path. `NetServiceConfig::get_peer_addr` → `get_peer_host_port`.

`bin/mosaic`'s TOML field shape is unchanged. The binary's `parse_host_port` validates the `host:port` shape without touching DNS, and rejects:
- bare IPv6 literals (require `[<addr>]:<port>`)
- empty hosts (`:9000`, `[]:9000`)
- missing ports (`operator.example.com`)

`PeerConfig::host_port()` emits IPv6 in bracketed form so it round-trips through the parser cleanly.

## Test plan

- [x] `cargo test --release -p mosaic-net-svc-api -p mosaic-net-svc -p mosaic` — all pass
- [x] `cargo +nightly fmt --all`, `cargo clippy --workspace --all-targets -- -D warnings` — clean
- [x] New: `unresolvable_peer_hostname_does_not_block_startup` integration test using `.invalid` TLD (RFC 6761) — asserts startup completes <4s and `open_protocol_stream` fails-not-wedges
- [x] New: `parse_host_port_accepts_*` and `parse_host_port_rejects_*` unit tests on the binary's parser
- [x] New: `from_host_keeps_hostname_for_lazy_resolution`, `get_peer_host_port_works` unit tests on the API
- [ ] Manual: boot a testnet node with one peer's DNS unpublished, confirm no crash; publish DNS, confirm the peer connects without restart

## Notes to reviewers

- Re-resolution on every attempt means an extra `getaddrinfo` per reconnect cycle. The OS resolver caches; under steady state with a healthy upstream this is negligible.
- Pre-handshake peer-id prediction is only a hint — TLS authenticates the peer identity, so partial maps (some peers unresolvable at startup) are correct, just less informative for logs.
- Multi-A-record consistency across attempts isn't guaranteed (resolver order can vary). Pre-existing behavior; TLS authenticates regardless.

### DIFF
--- a/bin/mosaic/src/config.rs
+++ b/bin/mosaic/src/config.rs
@@ -284,10 +284,13 @@ impl PeerEntry {
     }
 
     pub(crate) fn to_peer_config(&self) -> Result<PeerConfig> {
-        Ok(PeerConfig::new(
-            self.peer_id()?,
-            parse_socket_addr(&self.addr)?,
-        ))
+        // Address is validated as `host:port` shape but **not** resolved here
+        // — resolution happens at every outbound connection attempt. This
+        // lets mosaic boot cleanly even when a peer's DNS hasn't been
+        // published yet, and lets operators rotate their A records without
+        // requiring restarts on the other side.
+        let (host, port) = parse_host_port(&self.addr)?;
+        Ok(PeerConfig::from_host(self.peer_id()?, host, port))
     }
 }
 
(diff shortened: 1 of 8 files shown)


bin/mosaic/src/config.rs
```diff
@@ -441,6 +444,44 @@ fn parse_socket_addr(value: &str) -> Result<SocketAddr> {
...
+        bail!("address `{value}` has empty host");
+    }
+    // Reject bare IPv6: a host portion that itself contains ':' means the
+    // user wrote `2001:db8::1:9000`, which rsplit_once would mis-parse
+    // (the last `:9000` segment is `0` not the intended port).
+    if host.contains(':') {
+        bail!(
+            "address `{value}` looks like a bare IPv6 literal; use bracketed form `[<addr>]:<port>`"
+        );
+    }
```
sapinb, line 478: nit: comment is ambiguous on what the issue this check is trying to prevent is

sapinb (approved):
ACK 66d3dab82873e0d896eb7a3951764c4520c3a661

AaronFeickert:
Does this close #291?

PR #295, opened 2026-06-28
job/scheduler,config: raise the light pool concurrency default to 512

Zk2u:
## Summary

This PR now contains one change: the light pool default concurrency goes from
32 to 512.

The action container reuse is removed from this branch, per review feedback.
`split_off(0)` allocates a new Vec on every call, so the only saving was the
growth reallocations, in a path that storage and crypto dominate. The
take/restore rule also spanned 6 call sites, with no test to catch a missed
restore.

## Why 512

A setup burst from one peer can queue about 360 light-pool actions. Three
peers can queue more than 1000 at the worst moment. The old default of 32 gave
no headroom. Under stalled receives, sends and acks from other peers waited
behind the stalled work.

Light-pool tasks park on async I/O. Each unused slot costs a few KB. The heavy
pool keeps its default of 8 because it runs CPU-bound work.

## Per-pool defaults

The two pools shared one serde field default, and `default_heavy_pool_section`
applied only when the whole `[job_scheduler.heavy]` table was absent. A config
that declared that table with only `threads` therefore took the light pool
value for `concurrency_per_worker`. That value was 32 before, and a plain bump
makes it 512.

`PoolSection` now keeps an absent key as `None`, and the owning pool supplies
its default: 1 thread and 512 concurrency for light, 2 and 8 for heavy. This
also corrects the same fault on `threads`, where a heavy table without the key
gave 1 instead of 2.

`validate` now rejects `concurrency_per_worker = 0`. A worker takes a permit
before it pulls a job, so a worker with no permits accepts jobs that never
run.

Every config in this repository sets both keys, so no in-tree behavior
changes.

## What changed

- `DEFAULT_POOL_CONCURRENCY` splits into per-pool constants. Light goes to
  512, heavy keeps 8.
- The 6 example configs and the 2 docker configs drop the explicit
  `concurrency_per_worker = 32` line, so they use the new default.
- The functional-test config factory and `docs/observability.md` follow the
  new default.
- Three tests cover the defaults: absent tables, partly-filled tables, and
  zero concurrency.

The TOML field stays optional. A config that sets the value keeps its value.

## Test plan

- [x] `cargo +nightly fmt --all --check` — clean
- [x] `cargo clippy --workspace --all-targets -- -D warnings` — clean
- [x] `cargo test --release -p mosaic -p mosaic-job-scheduler` — 12, 39 and 17 pass
- [ ] Watch light-pool queue depth at testnet scale


storopoli:
@coder review

sapinb:
Should split the 2 changes into separate PRs.

The concurrency bump is fine to merge.

On the action container: the take/restore dance only exists because the container moved into shared state behind a RefCell. A fresh Vec per call, like the current code, has no await problem to begin with. The restore-on-every-exit-path rule spans 6 call sites, is enforced by review alone and a missed restore fails no test and just silently turns the optimization off. And since split_off(0) returns a newly allocated Vec, every call still pays an alloc + copy. The only real saving is the growth reallocations, in a path dominated by storage and crypto. Unless benchmarks show this appreciably boosts performance, I would prefer to keep it simple.

AaronFeickert (approved):

PR #304, opened 2026-07-07
fix: validate adaptor chunk indices

AaronFeickert:
### Motivation

The garbler handler previously indexed a peer-controlled `AdaptorMsgChunk` index into a fixed-size array without bounds checking, allowing an authenticated peer to trigger an out-of-bounds panic (DoS). 

### Description

- Add an explicit upper-bound check before accessing the relevant array.
- Add a regression test that sends an out-of-range chunk index and asserts the handler fails closed, emits no actions, and does not record the invalid chunk.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_b_6a484c0cc30c832ca212f6b655fa0870)


Zk2u (approved):
ack

PR #307, opened 2026-07-08
feat: cooperative scheduler hints for garbling table transfers

### LINKED ISSUE #306: Cooperative scheduling: sender-declared hints for the light pool
## Problem

Under sustained load, `receive_garbling_table` on the evaluator side times out repeatedly waiting for the sender's bulk stream to arrive, even when both peers are healthy and connected. Reported multiple times, most recently with attempts=136 (~90 min of retry churn); restart was the only fix.

Root cause is **scheduling asymmetry**, not pool saturation:

- `ReceiveGarblingTable` runs on the light pool (I/O-bound, 1×512 concurrency).
- `TransferGarblingTable` runs on the garbling coordinator (memory-bound, hardcoded to 8 concurrent sessions because each holds ~1 GB of RAM).
- The two sides are decoupled. Under load the sender's TGT can sit queued behind other garbling work for tens of seconds while the receiver's expectation timeout fires (30s).
- When the receiver times out it drops its expectation and retries. The sender's late-opened bulk stream then arrives at "no expectation, dropping stream" (`handlers.rs:1552`) and gets discarded. Retry cycle wedges — auto-recovery is probabilistic and can take hours.

Bumping light-pool concurrency (PR #295) helps at the queue-depth end but doesn't fix the underlying race. The sender is intrinsically slow-and-scarce (memory-bound); the receiver is intrinsically fast-and-abundant. The RTT+timeout protocol pretends the reverse.

## Proposal: cooperative scheduler hints

Best-effort scheduler-to-scheduler advisory messages. When the garbling coordinator dequeues a `TransferGarblingTable` and is about to open the bulk stream, it fires a `SchedulerMessage::TransferStarting { commitment }` to the peer on a dedicated high-priority stream. The peer's scheduler promotes the matching `ReceiveGarblingTable` job to the front of the light-pool queue so the expectation is registered by the time the bulk stream lands.

### Contract

- Every job in the queue still has a bounded timeout — hints modify **priority within the queue**, not lifetime or admission. No indefinitely-deferred jobs.
- Hints are advisory. Delivery failure → FIFO ordering, same as pre-hint behaviour. No regression.
- Boost queue is capped (64 slots vs 512 light concurrency) — a misbehaving peer can't reorder more than a small fraction of the queue.
- Fully out-of-band from the SM state machine. STF sees nothing new; determinism and replay-safety preserved.
- Wire forward-compat: peers on old code just don't send/receive hints, everyone falls back to FIFO.

### Scope

Initial variant: `TransferStarting`. Structure supports adding more variants (adaptor chunks, challenge responses) later without wire-breaking.

### Alternatives considered

- **Just bump concurrency**: doesn't fix the race, only reduces its probability. #295 is a band-aid.
- **Longer receive timeout**: parks light-pool slots for longer under stall. Trades one problem for another.
- **Add per-peer fairness to the light pool**: doesn't help — the wait itself is what starves throughput.
- **Push model (sender pushes without RTT)**: cleaner but a protocol change and touches STF.

## Related

- #295 (interim mitigation, light pool 32 → 512)

Zk2u:
Closes #306.

## Summary

Adds scheduler-to-scheduler advisory messages so the garbling coordinator can tell the receiver's scheduler "I'm about to send you table X" — the receiver promotes the matching `ReceiveGarblingTable` job to the front of its light-pool queue before the bulk stream arrives. Fixes the `receive_garbling_table` timeout wedge described in the linked issue.

## Design

- `HintKey { peer, kind, payload }` — opaque scheduler-layer identifier, O(1) hashable.
- `SchedulerMessage` on a dedicated `StreamType::SchedulerHint` stream at `StreamPriority::SchedulerHint = 2` (above Ack). Fire-and-forget; encode/decode is wire-forward-compat via `UnknownTag`.
- `JobQueue` FIFO backing is now a slot-based intrusive doubly-linked list with a `HashMap<HintKey, SlotId>` index. Push at tail and hint-driven promotion are both O(1). Boost queue drains completely before FIFO on pop.
- `PoolConfig::max_boost_slots: Option<usize>` — `Some(64)` for light (~12.5% of 512 concurrency), `None` for heavy/garbling. Boost overflow drops the hint (job stays FIFO, no worse than pre-hint baseline).
- `JobScheduler` accepts an optional inbound hint receiver; main `select!` loop decodes `SchedulerMessage`, maps to `HintKey`, calls `light.apply_hint`.
- `EvaluatorAction::ReceiveGarblingTable(commitment)` submit site derives `HintKey::new(peer_id, TRANSFER_STARTING, commitment)`.
- Sender: `job/executors::garbler::begin_table_transfer` fires `TransferStarting` hint just before `open_bulk_sender`, with a 5ms `futures_timer::Delay` so the peer's scheduler has a chance to promote before the bulk stream lands.
- `BULK_OPEN_TIMEOUT`: 30s → 10s. The hint mechanism tightens the sender-ready-to-receiver-ready gap enough that the timeout no longer needs to cover the full garbling-coordinator queue latency.

## Contract summary

- Hints are advisory. Delivery failure → FIFO ordering. No regression.
- Every queued job still has a bounded timeout. No indefinite deferral.
- Boost queue capped at 64 slots — bounds the fraction of the queue any peer can reorder.
- STF sees nothing new. State machine remains deterministic and replay-safe.
- Wire forward-compat: peers on old code don't send/receive hints; peers on new code sending to old code have their hint stream rejected at the header reader (unknown 0x02 tag) — nothing wedges.

## Test plan

- [x] Queue unit tests (11): FIFO ordering, boost promotion with observable ordering, boost cap, priority-mode ignore, hint-index cleanup after pop/promote (whitebox), slot free-list reuse, multiple-promotions arrival ordering
- [x] Wire type unit tests (4): `SchedulerMessage` encode/decode roundtrip, unknown tag, truncated, trailing bytes
- [x] Integration test (`mosaic-net-client`): two-peer wire path — `send_hint` from peer A delivers `InboundHintStream` to peer B's `hint_streams()` with `peer = A` and a decodable payload
- [x] `cargo test --release --workspace --lib --tests` — all pass
- [x] `cargo +nightly fmt --all`, `cargo clippy --workspace --all-targets -- -D warnings` — clean
- [x] Adversarial review-loop caught one blocker (net-svc header reader was rejecting tag 0x02 before dispatch) — fixed
- [ ] Manual: two-peer testnet under sender saturation, verify receiver timeouts stop even with 10s `BULK_OPEN_TIMEOUT`

## Notes to reviewers

- **Blocker caught by review**: `spawn_stream_header_reader` in `net-svc/tasks.rs` had an inline `match buf[0]` predicting the payload size that hardcoded 0x00/0x01 only. Any new `StreamType` variant needs to be added there OR the switch replaced with `StreamHeader::decode`. I opted to add `0x02 => 0` with a comment pointing at the wire-format source of truth. Worth considering a follow-up that removes the duplication.
- **Sender delay of 5ms** is a magic number. Chosen because typical local-scheduler-tick latency is sub-ms and QUIC handles high-priority stream frames ahead of bulk — 5ms is generous. Under a very busy net-svc runtime it could be too short; a follow-up could measure empirically or add a receiver-side grace window for late-arriving expectations.
- **Peer identity** on inbound hints is the TLS-authenticated peer id (traced through the review). No spoofing surface.
- Integration test only covers the wire path; scheduler-layer promotion is covered by unit tests. A two-peer test that exercises the whole loop (submit → hint → promote → dequeue) would need a slow-garbling-coordinator scenario and is a fair follow-up.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2557,10 +2557,12 @@ version = "0.1.0"
 dependencies = [
  "ckt-fmtv5-types",
  "fasm",
+ "hex",
  "kanal",
  "monoio",
  "mosaic-cac-types",
  "mosaic-job-api",
+ "mosaic-net-client",
  "mosaic-net-svc-api",
  "parking_lot",
  "tracing",
--- a/bin/mosaic/src/config.rs
+++ b/bin/mosaic/src/config.rs
@@ -83,11 +83,13 @@ impl MosaicConfig {
                 threads: self.job_scheduler.light.threads,
                 concurrency_per_worker: self.job_scheduler.light.concurrency_per_worker,
                 priority_queue: false,
+                max_boost_slots: Some(64),
             },
             heavy: PoolConfig {
                 threads: self.job_scheduler.heavy.threads,
                 concurrency_per_worker: self.job_scheduler.heavy.concurrency_per_worker,
                 priority_queue: true,
+                max_boost_slots: None,
             },
             garbling: GarblingConfig {
                 worker_threads: self.job_scheduler.garbling.worker_threads,
--- a/bin/mosaic/src/main.rs
+++ b/bin/mosaic/src/main.rs
@@ -222,8 +222,12 @@ where
         table_store,
         config.circuit.path.clone(),
     );
-    let (job_scheduler, job_handle) =
-        JobScheduler::new(config.build_job_scheduler_config(), job_executor);
+    let hint_rx = net_client.handle().hint_streams().clone();
+    let (job_scheduler, job_handle) = JobScheduler::new(
+        config.build_job_scheduler_config(),
+        job_executor,
+        Some(hint_rx),
+    );
     let job_scheduler_controller = job_scheduler.run();
 
     let (sm_executor, sm_executor_handle) = SmExecutor::new(
--- a/crates/job/api/src/hint.rs
+++ b/crates/job/api/src/hint.rs
@@ -1,0 +1,59 @@
+//! Scheduler hint identifiers.
+//!
+//! Hints are best-effort scheduler-to-scheduler messages that let a sender
+//! advise a receiver's scheduler that a specific piece of work is about to
+//! arrive, so the receiver can promote the matching job to the front of its
+//! queue. See the cooperative-scheduling design doc for the rationale.
+//!
+//! The queue-layer treats [`HintKey`] as an opaque `Hash + Eq` identifier;
+//! it does not interpret the discriminant or payload. The wire-level
+//! `SchedulerMessage` types in `mosaic-net-client` are responsible for
+//! building `HintKey` values from typed hint variants.
+
+use mosaic_net_svc_api::PeerId;
+
+/// Identifier used to route a scheduler hint to a matching queued job.
+///
+/// A hint fires from one peer's scheduler at another; the `peer` field
+/// carries the identity of the *sender* (i.e. "peer X told me they're about
+/// to send Y"). The receiver's scheduler builds the same [`HintKey`] at
+/// job-submission time and stores it alongside the job so incoming hints
+/// can locate the corresponding entry in O(1).
+///
+/// `kind` discriminates the hint family (e.g. `HintKind::TRANSFER_STARTING`).
+/// `payload` is family-specific — typically a hash-like identifier such as
+/// a garbling-table commitment.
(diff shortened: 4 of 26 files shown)


sapinb (approved):
ACK 331798e398d7dbc452e6fdb2d78d696013dceab8

PR #309, opened 2026-07-08
bin/mosaic: document RPC as operator-only, warn on non-loopback bind

### LINKED ISSUE #305: Warn about unauthenticated RPC
All RPC calls are unauthenticated. This is an intentional design choice, but it should be made clear in documentation.

Zk2u:
Closes #305.

## Summary

The JSON-RPC server is unauthenticated by design; its intended reader is the operator's own bridge node on a trusted internal network. The trust model wasn't spelled out anywhere in the repo, which caused an avoidable back-and-forth over a Codex finding.

- Doc comment on \`RpcConfig\` and the \`rpc\` module explaining the trust model.
- Warning line above \`[rpc]\` in \`config.example.toml\`.
- \`WARN\` log in \`start_rpc_server\` whenever \`bind_addr\` is not a loopback address, so the assumption shows up in operator logs at startup rather than only in prose.

## Test plan

- [x] \`cargo +nightly fmt --all\`
- [x] \`cargo clippy -p mosaic --all-targets -- -D warnings\`
- [ ] CI green

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,7 +77,13 @@ Copy [`bin/mosaic/config/config.example.toml`](bin/mosaic/config/config.example.
 - `[[network.peers]]` — one entry per other operator (`peer_id_hex` + `addr`)
 - `storage.cluster_file` — path to your `fdb.cluster`
 - `table_store` — `local_filesystem` (set `root`) or `s3_compatible` (set `bucket`, `region`, credentials)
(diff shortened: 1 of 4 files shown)


storopoli (approved):
ACK 3bf09523e4fcfd80ffab20157fa2d2e62b9a72e4

AaronFeickert (approved):

PR #319, opened 2026-07-20
job/executors,scheduler: make garbling-coordinator jobs idempotent against SM state

Zk2u:
Closes the garbling-coordinator live-lock observed on testnet (`no sessions created, pending=8` looping every 500ms while real transfers starved).

## Problem

The garbler STF deliberately re-emits `TransferGarblingTable` whenever the evaluator re-requests a table, so duplicate jobs are expected. But once the evaluator had receipted the table — or the SM had moved past `TransferringGarblingTables` entirely — a queued duplicate failed session setup with `StorageUnavailable`, which the coordinator treats as transient and retries forever. The coordinator drains its retry list first, bounded by `max_concurrent`: with `max_concurrent` permanently-stale jobs, every pass batch is 100% stale retries and the submission channel is never drained. New legitimate work starves indefinitely.

The E8 evaluation path had the same defect via a different route: after a deposit consumed the setup (`SetupConsumed`), a stale evaluation job resolved `StorageUnavailable` forever. G3/E3 commitment jobs did not retry forever, but a stale duplicate ran a full garbling pass (up to ~43 GB of ciphertext generation) only for the STF to reject its completion as `duplicate_action`.

## Fix

Jobs are idempotent. Each of the four coordinator session types now classifies itself against the SM root state before setup, via a pure resolver function over `(Step, index/seed)`:

- **Proceed** — the SM is in the expected step and the per-circuit bitmap (`generated` / `verified` / `transferred` / `evaluated`) says the work is still needed.
- **`CircuitError::AlreadyComplete`** (new variant) — the bitmap says the work is recorded, or the SM has moved past the step (including `Aborted` / `SetupConsumed`). The coordinator logs at info and drops the job. No completion is sent; these completions are informational to the SM, which has already moved on. Verified: fasm has no persisted action ledger — restore is state-derived and only re-emits actions for bitmap bits that are false, so dropping a recorded job can never strand required work.
- **`StorageUnavailable`** — the SM hasn't reached the step yet (unreachable in practice since the STF only emits in-step and never moves backward; kept as defense, where retrying is correct).
- **`SetupFailed`** — unknown seed/index (programming error; coordinator drops at error level, unchanged).

Resolvers: `resolve_pending_transfer` and `resolve_pending_garbler_commitment` (garbler.rs), `resolve_pending_evaluator_commitment` and `resolve_pending_evaluator_evaluation` (evaluator.rs). The E3 index lookup mirrors the STF's `binary_search` over the sorted `opened_indices`.

## Testing

- Unit tests for every classification of all four resolvers (23 tests), including receipted/generated/verified/evaluated bitmap hits, step-advanced, `Aborted`, pre-step, unknown index/seed, and non-contiguous opened indices.
- Pass under both circuit configurations (default and `reduced-circuits` feature unification).
- `cargo test --release --workspace --lib --tests`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo +nightly fmt` all clean.

### DIFF
--- a/crates/job/api/src/lib.rs
+++ b/crates/job/api/src/lib.rs
@@ -88,6 +88,13 @@ pub enum CircuitError {
     ChunkFailed(String),
     /// Transient network or I/O failure during session setup — retryable.
     TransientFailure(String),
+    /// The state machine has already recorded this action's work as done
+    /// (e.g. a duplicate transfer job for a table the evaluator has
+    /// receipted, or the SM has moved past the step entirely). The job is
+    /// an idempotent success: drop it without retry. Distinct from
+    /// [`StorageUnavailable`](Self::StorageUnavailable), which means the
+    /// data is not written *yet* and retrying will eventually succeed.
+    AlreadyComplete,
 }
 
 /// A block of circuit gate data shared across concurrent sessions via [`Arc`].
--- a/crates/job/executors/src/evaluator.rs
+++ b/crates/job/executors/src/evaluator.rs
@@ -12,7 +12,9 @@ use ckt_gobble::{
 use mosaic_cac_types::{
     Adaptor, ChallengeIndices, CircuitInputShares, DepositAdaptors, GarblingTableCommitment,
     TableTransferReceiptMsg, TableTransferRequestMsg, WideLabelWirePolynomialCommitments,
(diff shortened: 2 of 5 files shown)


sapinb (approved):
ACK 9a86beb9135ef6eda0310a5a90df7df59b2413fd

PR #323, opened 2026-07-29
ci: publish macos-arm64 mosaic-peer-id alongside linux-amd64 bare-metal binaries (STR-4133)

madan-oss:
External operators run `mosaic-peer-id` on their workstations to derive their peer ID, and the pikachu operator-3 onboarding surfaced that only `dev-cli` has a macOS arm64 prebuilt — `mosaic-peer-id` had none (Rust binaries are platform-specific, so the linux-amd64 artifact cannot run on a Mac). Slack context: https://alpen-labs.slack.com/archives/C0B6DUC9DCK/p1785272490075379

## Changes to `publish-baremetal-binaries.yml`

- **New `build-macos` job** (`macos-15`, Apple Silicon): natively builds **only `mosaic-peer-id`** — it has no FoundationDB linkage, so no FDB install is needed; the mosaic server itself stays Linux-only (links libfdb_c). Same security model as the linux job: no id-token, restore-only cache, artifact upload by explicit path only.
- **Publish job** now takes both artifacts, asserts both platforms resolved the same commit (a branch ref could advance between checkouts), checksums in the clean job as before, and uploads per-platform prefixes:
  - `binaries/mosaic/<commit>/linux-amd64/` — `mosaic`, `mosaic-peer-id`
  - `binaries/mosaic/<commit>/macos-arm64/` — `mosaic-peer-id`
- **Write-once is now per-platform**: an already-complete platform prefix is skipped instead of failing the run, so this workflow can be re-dispatched for `376d1bef` (whose linux-amd64 set was published today, run 30426812162) to add the macos-arm64 binary next to it. Fails only if *nothing* was published.
- Ref input documented as full-SHA/branch/tag only — the first dispatch today failed at checkout on a short SHA (`actions/checkout` cannot fetch abbreviated SHAs).

## Note for reviewers

~~Today's publish run did not pause at the environment gate.~~ **Correction:** the `baremetal-binaries` environment *is* protected (required reviewer + `main`-only branch policy) and the publish run was explicitly approved — verified via the run approvals API. The security model described in the workflow header is enforced as written. Optional hardening for later: `prevent_self_review` is off with a single reviewer, so the dispatcher can approve their own publish; a second reviewer + `prevent_self_review: true` would give four-eyes.

Jira: STR-4133

### DIFF
--- a/.github/workflows/publish-baremetal-binaries.yml
+++ b/.github/workflows/publish-baremetal-binaries.yml
@@ -155,11 +155,8 @@ jobs:
 
   build-macos:
     name: Build macos-arm64 mosaic-peer-id (untrusted ref, no credentials)
-    # Pinned (not -latest) for the same runner-drift reason as the linux job.
-    # Apple Silicon runner; produces an arm64 binary for operator workstations.
-    # Only mosaic-peer-id is built here — it has no FoundationDB linkage, so no
-    # FDB client install is needed on this runner (see header: a macOS mosaic
-    # build is deliberately out of scope).
+    # Apple Silicon runner, pinned for the same runner-drift reason as the linux
+    # job. mosaic-peer-id has no FoundationDB linkage, so no FDB install here.
     runs-on: macos-15
     timeout-minutes: 90
     permissions:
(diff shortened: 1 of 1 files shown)


madan-oss:
Addressed the review in f2fa9fc:

**Platform coupling** — done as suggested: publish now requires only the linux build (`!cancelled() && needs.build.result == 'success'`); the macOS download/checksum/publish steps are conditional on `needs.build-macos.result == 'success'`, with a warning + re-dispatch note when degraded. The same-commit assertion also only runs when both built.

**Environment protection** — the PR body note was stale and has been corrected: the `baremetal-binaries` environment *is* configured (required reviewer + main-only deployment branch policy), and the earlier publish run was explicitly approved (verified via the run approvals API). The header comment is accurate as written.

**"Changed publish code has never executed"** — fair; partially derisked by executing the extracted upload-step shell locally against the real bucket where both `376d1bef` platform prefixes already carry completion markers, so it only performs read-only `list-objects` calls. Verified all three cold paths: per-platform marker skip (both platforms), the `HAVE_MACOS=false` degraded warning branch, and the all-skipped `exit 1` error path. The happy write path is the same `aws s3 cp` sequence as the current workflow, unchanged except for the function wrapper.

AaronFeickert (approved):

